PR #15, opened 2021-09-20
Add support for analogue hikvision cameras

druchavda:


plugins/hikvision/src/main.ts
```diff
@@ -55,7 +55,66 @@ class HikVisionCamera extends RtspSmartCamera implements Camera {
...
+
+    getAnalogueCameraUrl() {
+        const channel = this.getRtspChannel()
+        const url = this.getRtspUrl()
+        const params = this.getRtspUrlParams()
+
+        return `${url}/${channel}01/${params}`
+    }
+
+    getRtspUrlOverride() {
```
koush, line 112: you'll want to override getCosntructedUrl instead of this method. Then return the custom url if it is analogue, otherwise return super.getCosntructedUrl

koush: Changing this method will break users wanting to override the RTSP url setting if they dont want to use the constructed URL you automatically provide.

PR #40, opened 2021-11-04
Plugin for Synology Surveillance Station

IGx89:
Creating a PR for initial review. Let me know how it looks! Planning on working on motion detection functionality next.

CC @koush

### DIFF
--- a/plugins/synology-ss/src/main.ts
+++ b/plugins/synology-ss/src/main.ts
@@ -1,4 +1,5 @@
 import sdk, { ScryptedDeviceBase, DeviceProvider, Settings, Setting, ScryptedDeviceType, VideoCamera, MediaObject, Device, ScryptedInterface, Camera, MediaStreamOptions, PictureOptions } from "@scrypted/sdk";
+import { createInstanceableProviderPlugin, enableInstanceableProviderMode, isInstanceableProviderModeEnabled } from '../../../common/src/provider-plugin';
 import { recommendRebroadcast } from "../../rtsp/src/recommend";
 import {SynologyApiClient, SynologyCameraStream, SynologyCamera} from "./api/synology-api-client";
 
@@ -12,16 +13,7 @@ class SynologyCameraDevice extends ScryptedDeviceBase implements Camera, Setting
         super(nativeId);
         this.provider = provider;
 
-        this.streams = [
-            { ...camera.stream1, id: '1' },
-            { ...camera.stream2, id: '2' },
-            { ...camera.stream3, id: '3' },
-        ].filter(s => !!s.resolution);
-
-        this.motionDetected = false;
-        if (this.interfaces.includes(ScryptedInterface.BinarySensor)) {
-            this.binaryState = false;
-        }
+        this.streams = SynologyCameraDevice.identifyStreams(camera);
     }
 
     private getDefaultOrderedVideoStreamOptions(vsos: MediaStreamOptions[]) {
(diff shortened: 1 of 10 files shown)


plugins/synology-ss/src/main.ts
```diff
@@ -0,0 +1,283 @@
...
+        return ret;
+    }
+
+    async putSetting(key: string, value: string | number) {
+        this.storage.setItem(key, value.toString());
+        this.discoverDevices(0);
+    }
+}
+
+export default new SynologySurveillanceStation();
```
koush, line 283: Unifi was using something like:

```ts
export default createInstanceableProviderPlugin("Unifi Protect Application", nativeid => new UnifiProtect(nativeid));
```

But I think you had issues with that. Was that resolved? It allows creation of multiple syno stations (optionally).

IGx89: Yeah, that was resolved. I removed it for simplicity at first, but could easily add it back in if you'd recommend?

IGx89: Unifi was the only plugin I saw using it, so wasn't sure if it was a recommended approach or a custom thing just for Unifi

plugins/synology-ss/src/main.ts
```diff
@@ -0,0 +1,283 @@
...
+
+    constructor(provider: SynologySurveillanceStation, nativeId: string, camera: SynologyCamera) {
+        super(nativeId);
+        this.provider = provider;
+
+        this.streams = [
+            { ...camera.stream1, id: '1' },
+            { ...camera.stream2, id: '2' },
+            { ...camera.stream3, id: '3' },
+        ].filter(s => !!s.resolution);
```
koush, line 19: I don't understand what is going on here without seeing the JSON, but my guess is that Synology always exposes 3 streams, which may or may not actually be valid (ie, no resolution).

plugins/synology-ss/src/main.ts
```diff
@@ -0,0 +1,283 @@
...
+
+        this.motionDetected = false;
+        if (this.interfaces.includes(ScryptedInterface.BinarySensor)) {
```
koush, line 22: might wanna just remove this check, since I don't think Synology has doorbell devices (binary sensor is the doorbell button)

plugins/synology-ss/src/main.ts
koush: Yeah, it came up for Unifi, but I could/should add this to any "hub" based provider like Hue, etc.

plugins/synology-ss/src/main.ts
IGx89: stream1/2/3 could be empty, so this logic is simply the most concise way I found to both set id labels on them and put them all in an array.

koush:
Looks good. There's duplicated code between Unifi and this, but that's fine. I'll refactor it to consolidate code between the hub based camera providers, like I've done with standalone RTSP camera types.

IGx89:
Thanks! I should have time to make the above modifications tomorrow night.

koush:
published. do you have an npm account? I can give you publish rights on that.

IGx89:
Thank you! Yeah, my NPM username is mdlieder. I'm totally fine with doing things like this here too, with a PR and a publish by you when merged. Whatever you feel is best! Expecting a few more version bumps in the next couple months, then (hopefully?) a long stable time.

PR #54, opened 2021-11-18
synology-ss: add motion detected webhook

IGx89:
Eventually plan to automatically set up alerts in Surveillance Station using their API, but in the meantime this makes it possible by manually setting up alerts. Been testing it myself over the past week and works pretty well. Figuring out how three separate motion timeouts (Scrypted, Surveillance Station and on camera) interact is a bit hair-pulling though -- still unsure how best to coordinate them, or if it even matters.

### DIFF
--- a/plugins/synology-ss/package-lock.json
+++ b/plugins/synology-ss/package-lock.json
@@ -1,12 +1,12 @@
 {
    "name": "@scrypted/synology-ss",
-   "version": "0.0.2",
+   "version": "0.0.3",
    "lockfileVersion": 2,
    "requires": true,
    "packages": {
       "": {
          "name": "@scrypted/synology-ss",
-         "version": "0.0.2",
+         "version": "0.0.3",
          "license": "Apache",
          "dependencies": {
             "axios": "^0.24.0"
@@ -17,8 +17,7 @@
          }
       },
       "../../sdk": {
-         "name": "@scrypted/sdk",
-         "version": "0.0.93",
+         "version": "0.0.121",
          "dev": true,
          "license": "ISC",
          "dependencies": {
@@ -38,7 +37,10 @@
             "ncp": "^2.0.0",
             "raw-loader": "^4.0.2",
             "rimraf": "^3.0.2",
+            "stringify-object": "^3.3.0",
+            "tmp": "^0.2.1",
             "ts-loader": "^9.2.6",
+            "typedoc": "^0.22.8",
             "typescript-json-schema": "^0.50.1",
             "webpack": "^5.59.0"
          },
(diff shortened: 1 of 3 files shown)


koush:
> Figuring out how three separate motion timeouts (Scrypted, Surveillance Station and on camera) interact is a bit hair-pulling though -- still unsure how best to coordinate them, or if it even matters.

There's various devices that have sensor "trigger" events, but don't actually report them as a binary state. Using a user configurable timeout is fine. I generally just default to 30 seconds. There's no particularly good way to handle it beyond this unfortunately.

I'm not sure what you mean by three separate motion timeouts though?

IGx89:
> I'm not sure what you mean by three separate motion timeouts though?

Probably not explaining it properly. If there's constant motion, the camera has a setting to only send motion events every X seconds. Synology and Scrypted have the same setting. Would setting all three to 60 seconds cause HomeKit to get motion events every 60 seconds if there's continuous motion, or is their timing not quite in sync so it would change to every 120 seconds, or even 180. Doesn't really matter, but I like preciseness :).

> Is the webhook a trigger, or does it maybe POST the state as a payload?

It can be a GET or POST, with a customized content type and payload. Payload is static as far as I can tell though, so wouldn't be any help to us (unless, say, the webhook URL didn't contain the camera ID)

koush:
> Probably not explaining it properly. If there's constant motion, the camera has a setting to only send motion events every X seconds. Synology and Scrypted have the same setting. Would setting all three to 60 seconds cause HomeKit to get motion events every 60 seconds if there's continuous motion, or is their timing not quite in sync so it would change to every 120 seconds, or even 180. Doesn't really matter, but I like preciseness :).

Motion in Scrypted is a binary state. It doesn't need a continuous update of the state. Only when it is started and ended.

Since Synology keeps events on an interval, you can set a timeout to something slightly longer than that interval (maybe 2 * interval), and set `motionDetected` to false when that timer expires without any further events. It does not need to be accurate for when motion ends, Homekit does not use that value. Homekit only cares about the motion start.

IGx89:
@koush just discovered that the 0.0.3 package you published has the updated package.json, but dist/plugin.zip appears to be from 0.0.2. Also, FYI since 11/17 HomeKit motion alerts appear to be flipped: I get notifications on my phone when motionDetected turns to false. Fairly certain that's not a problem with my plugin, but maybe I need to recreate my Docker image. Best guess is that the autoupdater updated some plugin that broke it.

koush:
There was a bug that caused the state to be inverted in the UI. Should be fixed in core update.

koush:
Let me take a look, what you're mentioning sounds like a non-UI issue. I did a rebuild/publish to .4 as well.

koush:
This is a race condition in my code, will fix this tonight. Will likely require a docker update.

IGx89:
Thanks!

koush:
Should be fixed now

IGx89:
Works great now, thanks again!

PR #58, opened 2021-11-23
Ecobee plugin

billyzoellers:
Mostly complete with a few work in progress items:

- Support getCode/getToken from UI
- UI buttons to enable/disable separate humidify/fan control in HomeKit
- Cleanup extraneous logging into "debug" UI option

### DIFF
--- a/plugins/ecobee/src/main.ts
+++ b/plugins/ecobee/src/main.ts
@@ -57,10 +57,18 @@ class EcobeeThermostat extends ScryptedDeviceBase implements HumiditySensor, The
   /* initialconfig(): set initial device characteristics
    *
    */
-  initialconfig(): void {
+  initialconfig(data): void {
     this.temperatureUnit = TemperatureUnit.F
     var modes: ThermostatMode[] = [ThermostatMode.Cool, ThermostatMode.Heat, ThermostatMode.Auto, ThermostatMode.Off];
     this.thermostatAvailableModes = modes;
+
+    this.console.log(data);
+    // set device info
+    this.info = {
+      model: data.brand,
+      manufacturer: data.modelNumber,
+      serialNumber: data.identifier,
+    }
   }
 
   /*
@@ -445,7 +453,7 @@ class EcobeeController extends ScryptedDeviceBase implements DeviceProvider, Set
       if (!device) {
         device = new EcobeeThermostat(devices[i].identifier, this)
         this.devices.set(devices[i].identifier, device)
-        device.initialconfig()
+        device.initialconfig(devices[i])
       }
     }
 
--- a/plugins/ecobee/src/main.ts
+++ b/plugins/ecobee/src/main.ts
@@ -245,7 +245,31 @@ class EcobeeThermostat extends ScryptedDeviceBase implements HumiditySensor, The
 
   async turnOff(): Promise<void> {
     this.console.log(`fanOff`)
-    // TODO: would need to resume program
+    // resume program
+    // https://www.ecobee.com/home/developer/api/documentation/v1/functions/ResumeProgram.shtml
+    const data = {
+      selection: {
+        selectionType:"registered",
+        selectionMatch: this.nativeId,
+      },
+      functions: [
+        {
+          type:"resumeProgram",
+          params:{
+            resumeAll: "false",
+          }
+        }
+      ]
+    }
+
+    var resp = await this.provider.req('post', 'thermostat', data, { format: "json" })
+    if (resp.status.code == 0) {
+      this.console.log("fanOff success")
+      await this.reload();
+      return;
+    }
+
+    this.console.log(`fanOff failed: ${resp}`)
   }
 
   async turnOn(): Promise<void> {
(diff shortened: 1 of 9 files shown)


plugins/ecobee/src/main.ts
```diff
@@ -0,0 +1,649 @@
+import axios, { AxiosRequestConfig } from 'axios'
+import sdk, { Device, DeviceInformation, ScryptedDeviceBase, OnOff, DeviceProvider, ScryptedDeviceType, ThermostatMode, Thermometer, HumiditySensor, TemperatureSetting, Settings, Setting, ScryptedInterface, Refresh, TemperatureUnit, HumidityCommand, HumidityMode, HumiditySetting } from '@scrypted/sdk';
+const { deviceManager, log } = sdk;
+
+const API_RETRY = 2;
+
+// Convert Fahrenheit to Celsius, round to 2 decimal places
+function convertFtoC(f: number) {
```
koush, line 8: I'm guessing this is ecobee API that takes Fahrenheit values. That's very weird!

plugins/ecobee/src/main.ts
```diff
@@ -0,0 +1,649 @@
...
+  on: boolean;
+
+  constructor(nativeId: string, provider: EcobeeController, info: DeviceInformation) {
+    super(nativeId);
+    this.provider = provider;
+    this.revisionList = null;
+    this.info = info;
+
+    this.temperatureUnit = TemperatureUnit.F
+    var modes: ThermostatMode[] = [ThermostatMode.Cool, ThermostatMode.Heat, ThermostatMode.Auto, ThermostatMode.Off];
```
koush, line 72: I generally `const` everything, and change it to a `let` as necessary. var is defecto deprecated due to weird block scope behavior.

koush:
This looks good to merge. My only question is regarding the getToken call. That looks like an affordance for an oauth process. However, it seems you are using some other login process that generates a pin code that you then enter into some ecobee site? I'm not sure I follow, since I don't have one of these.

Scrypted supports OAuth:
https://developer.scrypted.app/#oauthclient

The callback URL is:
https://home.scrypted.app/web/oauth/callback

If you set that up, your Oauth callback will be delivered your plugin. This will require the Scrypted Cloud plugin.

You don't need to use this of course, I actually think this pin sideband method is preferable (if I'm understanding the login process correctly).

plugins/ecobee/src/main.ts
billyzoellers: The API actually uses integers to represent decimal numbers (in Fahrenheit). So 720 = 70degF, or 715 = 71.5degF. Weird for sure!

plugins/ecobee/src/main.ts
billyzoellers: Good call. I think I actually copied that from what is likely some way old code in scrypted-nest 😀

I think const definitely makes sense though, so I'll swap that.

billyzoellers:
Just for the sake of documentation, it's this process:
https://www.ecobee.com/home/developer/api/documentation/v1/auth/pin-api-authorization.shtml

So the process currently looks like:
- Get ClientId from ecobee dev portal, enter into plugin
- Call getCode
- Take code to Ecobee website, enter/accept it
- Call getToken to get access/refresh token

I want to implement that as a UI flow with a couple buttons, but it works from REPL currently.

koush:
Without UI, I'd suggest calling `this.log.a` to show an alert with instructions (also place these instructions in the this.console):
 * Show an alert if client id is missing
 * After putSetting on the client id, call getCode and save it.
 * Immediately cal getToken (it will fail)
   * f the plugin is ever reloaded, if getToken finds a code and fails, it will show the alert again.

billyzoellers:
Done. I had a couple `this.log.a`'s already, but not enough to complete the process without going into REPL. It should be possible the complete the process without dropping into REPL now.

koush:
Looking good. I don't see where/how the code is being saved, it may not be?
I also didn't notice this in the initial review, but if this behaves anything like other access/refresh tokens, the access token shouldn't be saved. They're short lived and in general I just trade the refresh for an access token on plugin load. Usually the token comes with a TTL value to let you know when to exchange it. Maybe ecobee is not returning short lived access tokens, you would know better than me.

I've used this library in the path to manage oauth2 requests:
https://www.npmjs.com/package/client-oauth2

koush:
If you've verified the alert + reload process works from no login, I can go ahead and merge this.

billyzoellers:
I tested the alert/reload process a few times and ended up cleaning up places that I was keeping something in storage and as a property for no reason. Also stopped storing the access token. So, the setup process is confirmed.

I'll add client-oauth2 to my list to check out for the next iteration. That might let me cleanup some of the code.

koush:
I've published the plugin on npm, take a look. Do you have an npmjs account? I can give you publish access to this plugin as well.

koush:
My publish script:

```
# in ecobee directory

# increase revision by 0.0.1
npm version patch

# publish it to npm
npm publish --access public

# add and commit the changed package.json
git add . && git commit -m 'ecobee: publish'
```

PR #60, opened 2021-11-24
Added menu item for setting additional audio parameters

systemx-xx:


koush:
Thanks for the pull request. I think what is happening here is that the bitstream filter is added before the mpegts output, similar to how it is added before the fragmented mp4 output. I don't think an option is necessary-- it should be there by default. I'll need to test that nothing breaks, then I'll add this flag by default.

koush:
the bsf adts to aac is now done automatically, this should be resolved.

PR #75, opened 2021-12-04
github action: publish to ghcr and dockerhub

ctso:
This change causes the Github docker workflow to publish the image to both Dockerhub and [Github Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry).

Why? Dockerhub rate limits downloads unless you pay them and authenticate with Dockerhub.  This can be problematic for people with larger setups that don't want to pay up to Dockerhub, like myself.

I suspect this will "just work".  I don't recall having to setup any permissions magic on my repositories that I've done this with.  I believe you get a `secrets.GITHUB_TOKEN` secret for free that has scoped access to this repo.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -22,17 +22,18 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
         
-      - name: Log in to Docker Hub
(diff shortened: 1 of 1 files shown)


koush:
yep Just Worked

PR #99, opened 2022-01-02
Support for old Hikvision NVR by suppressing unsupported requests and limiting number of alert streams

cfeda:
As per discussion on Discord, with the older Hikvision NVR (DS-7608NI-E2/8P) the system can stop processing new requests when 1) receiving requests it doesn't support (/ISAPI/Streaming/channels and Streaming/channels/CAM/capabilities) and 2) when receiving too many concurrent notification/alertStream requests.
These changes address this by:
1) Checking the deviceModel of the NVR and bypassing the unsupported requests (using the existing defaults)
2) Finishing the support for using a shared HikVisionCameraAPI with a single request stream in listenEvents().

### DIFF
--- a/plugins/hikvision/src/hikvision-camera-api.ts
+++ b/plugins/hikvision/src/hikvision-camera-api.ts
@@ -1,5 +1,6 @@
 import { Readable } from 'stream';
 import AxiosDigestAuth from '@koush/axios-digest-auth';
+import { EventEmitter } from "stream";
 
 export enum HikVisionCameraEvent {
     MotionDetected = "<eventType>VMD</eventType>",
@@ -24,6 +25,8 @@ export interface HikVisionCameraStreamSetup {
 
 export class HikVisionCameraAPI {
     digestAuth: AxiosDigestAuth;
+    deviceModel : Promise<string>;
+    listenerPromise : Promise<EventEmitter>;
 
     constructor(public ip: string, username: string, password: string, public channel: string, public console: Console) {
         this.digestAuth = new AxiosDigestAuth({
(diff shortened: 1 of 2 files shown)


plugins/hikvision/src/main.ts
```diff
@@ -26,13 +26,20 @@ class HikVisionCamera extends RtspSmartCamera implements Camera {
...
                 events.on('error', e => ret.emit('error', e));
                 events.on('event', (event: HikVisionCameraEvent, cameraNumber: string) => {
-                    // if (this.getRtspChannel() && cameraNumber !== this.getCameraNumber()) {
-                    //     return;
-                    // }
                     if (event === HikVisionCameraEvent.MotionDetected
                         || event === HikVisionCameraEvent.LineDetection
                         || event === HikVisionCameraEvent.FieldDetection) {
-                        this.motionDetected = true;
+                        if (this.getRtspChannel() && cameraNumber !== this.getCameraNumber()) {
```
koush, line 32: is there a reason you uncommented this filter? there exists a filter in hikvision-api.ts, is that not working?

plugins/hikvision/src/hikvision-camera-api.ts
```diff
@@ -32,7 +35,42 @@ export class HikVisionCameraAPI {
...
+        if (!this.deviceModel) {
+            this.deviceModel = new Promise(async (resolve, reject) => {
+                try {
+                    const response = await this.digestAuth.request({
+                        method: "GET",
+                        responseType: 'text',
+                        url: `http://${this.ip}/ISAPI/System/deviceInfo`,
+                    });
+                    const deviceModel = response.data.match(/>(.*?)<\/model>/)?.[1];
+                    resolve(deviceModel);
```
koush, line 48: I think another additional safe change could be made here: store this result in `this.storage.setItem('model', deviceModel)` and load it later. Can bypass this request forever once it succeeds. The model will also never change, obviously.

plugins/hikvision/src/main.ts
```diff
@@ -26,13 +26,20 @@ class HikVisionCamera extends RtspSmartCamera implements Camera {
...
+                            // this.console.error(`### Skipping motion event ${cameraNumber} != ${this.getCameraNumber()}`);
+                            return;
+                        }
+        
+                        // this.console.error('### Detected motion, camera: ', cameraNumber);
+                        const prevMotion = this.motionDetected;
+                        if (!prevMotion) { 
+                            this.motionDetected = true;
```
koush, line 40: Is there a reason for this? It shouldn't be necessary. Scrypted will automatically denoise no-op state changes. But it does log that the state was updated at a specific time.

plugins/hikvision/src/main.ts
```diff
@@ -26,13 +26,20 @@ class HikVisionCamera extends RtspSmartCamera implements Camera {
...
+                            this.console.log(`motionDetected, camera ${this.getCameraNumber()}`);
```
koush, line 41: This is unnecessary, as all state changes are automatically logged in the Events tab: 

image

koush:
Looks good, I'm a little hesitant to merge due to the filter change you made. Was there a reason for that?

plugins/hikvision/src/hikvision-camera-api.ts
cfeda: I see.  Is this.storage uniqued to ip+user (since a user could have multiple DVRs)?

plugins/hikvision/src/main.ts
cfeda: I did it just so a log event would be emitted only when an actual change occurred.  (I was getting overwhelmed with logs…)

plugins/hikvision/src/hikvision-camera-api.ts
koush: this.storage is unique per device, yes.

plugins/hikvision/src/main.ts
koush: > I did it just so a log event would be emitted only when an actual change occurred. (I was getting overwhelmed with logs…)

I see, I'll remove it then. I prefer to have those events, as sent from the device, even if they're duplicating the state. I can denoise it in the event log viewer.

PR #119, opened 2022-01-16
Adding support for Dahua doorbells

kvazibox:
There are two areas where Dahua doorbells (at least the model that I have tested, VTO2202F-P-S2) behave differently:

1. Event codes are used differently. More specifically, AmcrestEvent.TalkPulse should be ignored for Dahua doorbells. Also, other 3 new events were added.
2. Intercom also works slightly differently, we need to accumulate and send segments when they are over 1KB in size. 

I have added a setting for Dahua Doorbells to activate this different behaviour, however, I need your help figuring out how to better manage and present that setting. Right now if a user enables the Dahua Doorbell setting but disables the Amcrest Doorbell setting, it will disable the binary sensor which is obviously not what we want.

### DIFF
--- a/plugins/amcrest/src/amcrest-api.ts
+++ b/plugins/amcrest/src/amcrest-api.ts
@@ -14,6 +14,9 @@ export enum AmcrestEvent {
     AlarmIPCStop = "Code=AlarmIPC;action=Stop",
     PhoneCallDetectStart = "Code=PhoneCallDetect;action=Start",
     PhoneCallDetectStop = "Code=PhoneCallDetect;action=Stop",
+    DahuaTalkInvite = "Code=CallNoAnswered;action=Start",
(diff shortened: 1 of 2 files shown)


koush:
I appreciate that you put the flags behind a dahua doorbell. But I think the distinction that is a dahua doorbell may be unnecessary if the audio streaming method that is hidden behind the flag also works on existing amcrest doorbells I(it should). I will confirm and probably remove that flag.

kvazibox:
Well, the biggest problem with removing the flag is the conflicting event code AmcrestEvent.TalkPulse. It has to be ignored for Dahua doorbells (at least in my testing).

The new Dahua two way audio implementation should *probably* be fine for Amcrest doorbells as well but it is better to ask the existing users to test that.

PR #121, opened 2022-01-16
Filter out non-human voice frequency audio. 

vandalon:
Filter out non human voice frequency to improve conversation quality.


koush:
As I mentioned in discord, putting filters in the ffmpeg input arguments is a hack, and I can't accept that change until I get the plumbing correctly. This will break other things down stream.

koush:
I'll leave this open to track.

PR #127, opened 2022-01-18
Add ZWave Security Keys and Barrier Operator Device Support

marcptrs:
- Added support to specify security keys according to here: [Security S2](https://zwave-js.github.io/node-zwave-js/#/getting-started/security-s2)
- Added support for Barrier Operator devices such as garage doors

Main issue is that the ZWave plugin will fail to load during the first run if you have the ZWave plugin installed as it will be attempting to convert the existing base64 storage value as hex. I was able to get my ZWave up and running after deleting the plugin, closing my browser, and then going through the setup again. Would be nice to have a way to convert an existing setup to a hex string value, but I'm not sure what the best way to do this would be. I suppose you can test if the existing value ends in "==" and go from there?

I've confirmed my garage door is added as a new device and I can open/close it with the appropriate state being reported. Super excited!

### DIFF
--- a/plugins/zwave/package.json
+++ b/plugins/zwave/package.json
@@ -1,6 +1,6 @@
 {
    "name": "@scrypted/zwave",
-   "version": "0.0.30",
+   "version": "0.0.31",
    "description": "Z-Wave USB Controller for Scrypted",
    "author": "Scrypted",
    "license": "Apache",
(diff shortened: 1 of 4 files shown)


koush:
oh wow thanks so much! didn't think anyone was using this for anything but cameras... :) I run ~80 zwave devices myself.

koush:
published

koush:
> Main issue is that the ZWave plugin will fail to load during the first run if you have the ZWave plugin installed as it will be attempting to convert the existing base64 storage value as hex.

Hm, I'll run through the first run/import again for bugs on burner setup... it's been a while.

koush:
I also need to fix the entry/lock to not be booleans: ie, have transient/target/error states.

koush:
Added a network key migration before updating to your patch, and it worked great. de39155

PR #131, opened 2022-01-20
Multiple Enhancments

donavanbecker:
Added option to only show specific locationIds
Added option to set cameraDingsPollingSeconds
Added firmware info
Added serialNumber info.
Added Intercom to all devices.

### DIFF
--- a/plugins/ring/src/main.ts
+++ b/plugins/ring/src/main.ts
@@ -151,6 +151,16 @@ class RingPlugin extends ScryptedDeviceBase implements DeviceProvider, DeviceDis
         },
         refreshToken: {
             hide: true,
+        },
+        locationIds: {
+            title: 'Location ID',
+            description: 'Optional: If supplied will on show this locationID.',
+            
+        },
+        cameraDingsPollingSeconds: {
+            title: 'Camera Event Polling',
+            type: 'number',
+            description: 'Optional: Camera motion and ding events does not update live and must be fetched periodically.  By default this is done every 2 seconds',
         }
     }, this.storage);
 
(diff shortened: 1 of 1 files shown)


plugins/ring/src/main.ts
```diff
@@ -151,6 +151,16 @@ class RingPlugin extends ScryptedDeviceBase implements DeviceProvider, DeviceDis
         },
         refreshToken: {
             hide: true,
+        },
+        locationIds: {
+            title: 'Location ID',
+            description: 'Optional: If supplied will on show this locationID.',
```
koush, line 157: what does location id do?

donavanbecker: So when a locationId is supplied, it will then only show devices for that locationId

Before:Screen Shot 2022-01-19 at 10 42 09 PM

After: Screen Shot 2022-01-19 at 10 42 32 PM

koush: What is the reason to do this? Generally it's fine to have everything imported, and then filter which services (Homekit, google home) get which synced devices.

donavanbecker: If someone accidentally checks a camera, then it will get added again. Where this they can manually set the locationIds of the cameras.

Like for my use case. I will never use the cameras I removed using this. Since they are from a shared account. Those cameras wouldn't be part of my house, but someone else's.

koush: I think it is fine to have this code, for debugging. I am going. To hide the setting. You can still access it by changing thread storage value directly.
I don’t think this is Useful to normal users who may fill this field out and then not see any devices.

PR #145, opened 2022-01-27
Support custom API endpoints in SDM plugin

andrewheard:
Allows the user to customize the endpoints used for authorization and
accessing the API in the google-device-access plugin by specifying
hostnames in storage. These can be configured using the
`authorizationHostname` and `apiHostname` keys, respectively.


koush:
huh, what's this used for?

andrewheard:
Thanks for merging this in. I've got a bunch of old devices stuck in a beta environment and now I won't have to keep a separate copy of your plugin. Seems to work just the same other than the hostnames.

PR #152, opened 2022-01-29
Register as a service via launchd on macOS

adrum:
This sets up scrypted to run under launchd as the currently logged in user. It will run at load, and keep alive in case it crashes.

It also will automatically detect the correct bin folder from brew, so it can run on Intel and Apple Silicon.

### DIFF
--- a/docker/install-scrypted-dependencies-mac.sh
+++ b/docker/install-scrypted-dependencies-mac.sh
@@ -69,17 +69,14 @@ RUN cat << EOF | tee ~/Library/LaunchAgents/com.scrypted.server.plist
 EOF
 
 RUN launchctl load ~/Library/LaunchAgents/com.scrypted.server.plist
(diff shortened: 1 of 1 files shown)


docker/install-scrypted-dependencies-mac.sh
```diff
@@ -30,15 +30,54 @@ RUN pip3 install aiofiles debugpy typing_extensions typing opencv-python
...
+<dict>
+    <key>RunAtLoad</key>
+        <true/>
+    <key>KeepAlive</key>
+        <true/>
+    <key>Label</key>
+        <string>com.scrypted.server</string>
+    <key>ProgramArguments</key>
+        <array>
+             <string>$(brew --prefix)/bin/npx</string>
```
koush, line 47: would use $(which npx) I think, or $0? Probably the prior. Should probably also do all sub shells first, with the RUN, to guarantee they succeed before creating a potentially busted launchd.

PR #158, opened 2022-01-30
QNAP Installation Guide

lastowl:
Added an installation guide for QNAP NAS installations via docker/Container Station


koush:
Excellent thank you!

PR #208, opened 2022-03-30
Arlo improvements

bjia56:
Changes:
 - Use storage API for settings and cleaned up shims that are now irrelevant due to an updated Python SDK
 - Use `self.print` where possible
 - Skip cameras that do not have a valid basestation, as they won't work (Arlo requires a basestation to subscribe to)
 - Make camera `firmwareVersion` optional


plugins/arlo/src/arlo_plugin/provider.py
```diff
@@ -135,23 +103,28 @@ async def discoverDevices(self, duration=0):
...
+            if camera["deviceId"] != camera["parentId"] and camera["parentId"] not in self.arlo_basestations:
+                self.print(f"Skipping camera {camera['deviceId']} because its basestation was not found")
+                continue
+
             device = {
                 "info": {
                     "model": f"{camera['properties']['modelId']} ({camera['properties']['hwVersion']})",
                     "manufacturer": "Arlo",
-                    "firmware": camera["firmwareVersion"],
+                    "firmware": camera.get("firmwareVersion"),
```
koush, line 127: this will throw if arlo's payload doesn't have it for some random reason

PR #223, opened 2022-04-15
sendgrid: initial implementation

bjia56:
Opening this PR to upstream the plugin, if you think it'd be useful enough to include in the main repo. 😃 

Getting started with SendGrid pretty much requires the user to own a domain name and be able to configure DNS records for domain authentication. Per SendGrid docs, it doesn't look like using public email domains like gmail, yahoo, etc. will actually deliver emails correctly.

### DIFF
--- a/plugins/sendgrid/src/main.ts
+++ b/plugins/sendgrid/src/main.ts
@@ -35,10 +35,6 @@ class SendGridProvider extends ScryptedDeviceBase implements Notifier, Settings 
             return
         }
 
-        if (this.sendgridClient) {
-            return
-        }
-
         this.sendgridClient = new MailService();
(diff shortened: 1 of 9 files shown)


plugins/sendgrid/src/main.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        await this.sendgridClient.send(msg);
+        this.console.info(`Email sent to ${this.to()}`)
+    }
+};
+
+const provider = new SendGridProvider();
+
+export default provider;
```
koush, line 111: you can return the class in Typescript plugins now, don't need to instantiate it anymore.

koush:
```typescript
export default SendGridProvider;
```

PR #236, opened 2022-05-03
homekit: add support for Air Quality Sensor

billyzoellers:
Add support for Air Quality Sensor with Air Quality, VOC, and PM2.5 values


plugins/homekit/src/types/sensor.ts
```diff
@@ -75,6 +94,16 @@ addSupportedType({
                 () => !!device.flooded);
         }
 
+        if (device.interfaces.includes(ScryptedInterface.AirQualitySensor)) {
```
koush, line 97: check for the existence of each interface

PR #239, opened 2022-05-05
zwave: allow manual field entry

billyzoellers:
- Updating zwave-js to 9.2.1
- Support devices that have both "choices" and allow values to be filled in

### DIFF
--- a/plugins/zwave/src/CommandClasses/SettingsToConfiguration.ts
+++ b/plugins/zwave/src/CommandClasses/SettingsToConfiguration.ts
@@ -21,6 +21,8 @@ export class SettingsToConfiguration extends ZwaveDeviceBase implements Settings
             setting.description = metadata.description;
             if ((metadata as any).states)
                 setting.choices = Object.values((metadata as any).states);
+                if ((metadata as any).allowManualEntry)
+                    setting.combobox = true;
             const value = node.getValue(valueId) as any;
             setting.value = setting.choices?.[value] || value;
             settings.push(setting);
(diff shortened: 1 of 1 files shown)


plugins/zwave/src/CommandClasses/SettingsToConfiguration.ts
```diff
@@ -38,7 +38,10 @@ export class SettingsToConfiguration extends ZwaveDeviceBase implements Settings
         const node = this.instance.getNodeUnsafe();
         const metadata = node.getValueMetadata(valueId);
         if ((metadata as any).states) {
-            value = Object.entries((metadata as any).states).find(([, v]) => v === value)?.[0];
+            const state_value = Object.entries((metadata as any).states).find(([, v]) => v === value)?.[0];
+            if (state_value)
+                Object.entries((metadata as any).states).find(([, v]) => v === value)?.[0];
+            this.console.log(value)
```
koush, line 44: should remove this logging I think

plugins/zwave/src/CommandClasses/SettingsToConfiguration.ts
```diff
@@ -14,13 +14,13 @@ export class SettingsToConfiguration extends ZwaveDeviceBase implements Settings
...
         for (var valueId of values) {
             const metadata = node.getValueMetadata(valueId);
-
             let setting: Setting = {};
             setting.key = valueId.property.toString();
             setting.title = metadata.label;
             setting.description = metadata.description;
             if ((metadata as any).states)
                 setting.choices = Object.values((metadata as any).states);
+                setting.combobox = true;
```
koush, line 23: indent? was this meant to be under the if or always applied?

koush: why would you want free text entry? I'm not sure that's valid

billyzoellers: This was intentional. i.e. "make it a combobox if there are states(choices)"

billyzoellers:
@koush fixed to use the allowManualEntry field

PR #241, opened 2022-05-07
homekit: map EntrySensor to homekit ContactSensor

billyzoellers:
Door/open close sensor in ZWave is a ScryptedInterface.EntrySensor which was not supported by homekit


koush:
doesn't it also expose a binarysensor which is exposed? usually zwave does both.

billyzoellers:
In this case, no. It's an Aeotec Refressed Door Sensor 7

```
> device.interfaces
[
  'Battery',
  'EntrySensor',
  'Online',
  'PowerSensor',
  'Refresh',
  'Settings',
  'mixin:@scrypted/homekit'
]
```

koush:
Multiple contact sensors will need unique names to prevent HAP service clash. See how I handled the audio sensor type in the same file.

billyzoellers:
Updated to avoid that. Do your door sensors present both BinarySensor and EntrySensor? If so, this probably causes a duplicate in HomeKit.

koush:
I think they do. You could add it only if a binary sensor isn’t found.

billyzoellers:
Should be good now.

PR #242, opened 2022-05-08
Elgato Key Light Plugin

3nth:
a plugin to support Elgato Lights. Tested with a Key Light Air.


koush:
This is awesome, thank you! There's nothing preventing you from publishing this as an add-on yourself (package renamed  to scrypted-elgato or something outside of the npm org). It will still be available for all users.

If you do want to submit it as a plugin in the scrypted npm org, create a separate/new repository for this plugin. I'm in the process of moving non-camera repositories out of this mono repo.  The plugin can then be published at: https://github.com/scryptedapp and would be available as @scrypted/elgato.

Either way works for me, so it's up to you on how you want to publish this.

plugins/elgato/package.json
```diff
@@ -0,0 +1,42 @@
...
+   "scrypted": {
+      "name": "Elgato Controller",
+      "singleInstance": true,
+      "type": "DeviceProvider",
+      "interfaces": [
+         "DeviceProvider"
+      ]
+   },
+   "devDependencies": {
+      "@scrypted/sdk": "file:../../sdk",
```
koush, line 33: when you create your own repository, I think the only thing you'll need to change is this to

```
      "@scrypted/sdk": "^0.0.145"
```

plugins/elgato/package.json
```diff
@@ -0,0 +1,42 @@
...
+      "@types/bonjour": "^3.5.10",
```
koush, line 34: would suggest using ciao (what HomeKit plugin uses) because bonjour is unmaintained (last release was 6 years ago)

plugins/elgato/package.json
```diff
@@ -0,0 +1,42 @@
...
+      "@types/node": "^16.7.1",
+      "@types/request-promise": "^4.1.48"
+   },
+   "dependencies": {
+      "bonjour": "^3.5.0",
```
koush, line 39: would suggest using ciao (what HomeKit plugin uses) because bonjour is unmaintained (last release was 6 years ago)

plugins/elgato/package.json
```diff
@@ -0,0 +1,42 @@
...
+      "request-promise": "^4.2.6"
```
koush, line 40: request-promise has been deprecated (and is quite bulky since it uses bluebird instead native javascript promise). axios is a nice alternative that is an easy switch.

plugins/elgato/src/main.ts
```diff
@@ -0,0 +1,117 @@
...
+  async newLight(light: KeyLight) {
+    await light.refresh();
+    var info = {
+      name: light.info.displayName,
+      nativeId: light.info.serialNumber,
+      interfaces: ['OnOff', 'Brightness', 'ColorSettingTemperature', 'Refresh'],
+      type: ScryptedDeviceType.Light,
+    };
+
+    deviceManager.onDeviceDiscovered(info);
```
koush, line 97: `await` this call before creating the light

plugins/elgato/src/main.ts
```diff
@@ -0,0 +1,117 @@
...
+    this.lights[light.info.serialNumber] = new ElgatoDevice(light, info);
+  }
+
+  async discoverDevices(duration: number) {
+    const browser = bonjour().find({ type: 'elg' });
+    log.i(`discoverDevices...`)
+    browser.on('up', service => {
+        log.i(`found light: ${service.name}`)
+        let newLight = new KeyLight(service['referer'].address, service.port, service.name);
```
koush, line 106: I think you should use an id rather than address here, since dhcp can/will change the address. then cache the address.

plugins/elgato/src/lights.ts
```diff
@@ -0,0 +1,123 @@
...
+    ip: string;
+    port: number;
+    name: string;
+    settings?: KeyLightSettings;
+    info?: KeyLightInfo;
+    options?: KeyLightOptions;
+
+    constructor(ip: string, port: number, name: string)
+    {
+        this.ip = ip;
```
koush, line 13: call the super(nativeId) with a unique id for this light.

plugins/elgato/package.json
```diff
@@ -0,0 +1,42 @@
...
+   },
+   "keywords": [
+      "elgato",
+      "keylight",
+      "scrypted",
+      "plugin"
+   ],
+   "scrypted": {
+      "name": "Elgato Controller",
+      "singleInstance": true,
```
koush, line 26: can remove this, I actually have to remove it from the typescript sample too.

plugins/elgato/package.json
koush: on second thought, I'm unsure if ciao does discovery.

3nth:
@koush cool. cool cool cool. I will check out all the things mentioned.

I started off copying the lifx plugin, so it is basically almost the same. 

I can also contribute to some plugin developer doc as I figure it all out.

thanks for the spiffy system!

plugins/elgato/package.json
3nth: I take it that this would be a good reference for setting up as a separate repo?
https://github.com/scryptedapp/sendgrid

plugins/elgato/package.json
3nth: axios is quite nice!

plugins/elgato/package.json
3nth: yeah, not as far as I can tell, but someone did pick fork bonjour and is going typescript on it:
https://github.com/onlxltd/bonjour-service
https://www.npmjs.com/package/bonjour-service

plugins/elgato/package.json
3nth: also need to add a dev dependency on `webpack-bundle-analyzer` but otherwise is working after splitting it out:
https://github.com/3nth/scrypted-elgato

3nth:
thanks for the review. I made all those changes, split into a separate repo and got it published! very cool.

only thing I cannot figure is how to log to the plugin console.

Also, does the device discovery run periodically? As you mentioned, the IP could change. This is storing devices by serial number and my assumption is a refresh would simply update the existing device with the new IP, but I don't know that for sure....

koush:
this.console is available in ScryptedDeviceBase

koush:
device discovery is not run periodically at the moment, I can add a ui button to do it manually. Various upstream services like Alexa willl trigger discovery though when you ask it to search for devices.

plugins/elgato/src/main.ts
```diff
@@ -0,0 +1,117 @@
...
+  constructor(light: KeyLight, device: Device) {
+    super(device.nativeId);
+    this.light = light;
+    this.device = device;
+
+    this.refresher = (err) => this.refresh();
+
+    // schedule a refresh. not doing it immediately, to allow the device to be reported
+    // by sync first.
+    setImmediate(() => this.refresh());
```
koush, line 21: Since you have the await on the deviceManger now you can probably do this sync

plugins/elgato/src/main.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  getDevice(id) {
+    return this.lights[id];
+  }
+
+  async newLight(light: KeyLight) {
+    await light.refresh();
+    var info = {
+      name: light.info.displayName,
+      nativeId: light.info.serialNumber,
```
koush, line 92: Can use ScryptedInterface.OnOff for typed values which resolve to strings. I will update the LIFX sample too.

3nth:
Okay, this is working like a champ now:
https://github.com/3nth/scrypted-elgato

If light IPs changed a click on "Reload Plugin" should find and update if I understand correct: same nativeId (serial number), but the ElgatoController will have updated info.

PR #244, opened 2022-05-12
homekit: support IntrusionSensor as ContactSensor service

kevinwmerritt:
Support IntrusionSensor in Homekit.

Context: https://github.com/koush/scrypted/blob/main/plugins/zwave/src/CommandClasses/index.ts#L88

The Z-Wave Home Security CC maps to IntrusionSensor.


koush:
Usually the intrusion sensors are paired with other sensors, like motion, etc. Those Services in HAP have a StatusTampered characteristic. Would it be more appropriate to use that characteristic? What other command classes does your intrusion sensor advertise?

kevinwmerritt:
I was going to suggest that separately there be a Tamper (isTampered) interface added to the SDK.

The device in particular is a Ring Outdoor Contact sensor. It supports a few different tamper notifications such as magnetic interference and case removed. Any of these tamper notifications could be mapped to the isTampered state and then on to the StatusTampered characteristic in HAP.

The main contact sensor notification is Home Security state: Intrusion and state: Intrusion cleared/idle.
Link to manual: https://support.ring.com/hc/en-us/article_attachments/4414661689748/Ring_Alarm_Outdoor_Contact_Sensor_Zwave.pdf

PR #257, opened 2022-05-23
Fix incorrect assumption that LIBX264_ENCODER_TITLE is always set

joshkulesza:
LIBX264_ENCODER_TITLE is not set under any condition where the is platform is unrecognised.

FreeBSD is not mapped in `common/src/ffmpeg-hardware-acceleration.ts` and as a result the rebroadcast plugin is unable to start due to `getH264EncoderArgs()[LIBX264_ENCODER_TITLE]` being undefined when `join` is called


koush:
it should always be set though, it's a constant. something else is going wrong.

koush:
may be some strange issue with side effects or something.
I'll use `getDebugModeH264EncoderArgs` instead.

koush:
baef74c7

koush:
and ffac90c5

PR #258, opened 2022-05-23
homekit: airquality sensor updates

billyzoellers:
bugfix: airquality sub-sensor displayed when it shouldn't be

support air quality sensor as part of thermostat type


plugins/homekit/src/types/sensor.ts
```diff
@@ -104,10 +104,15 @@ addSupportedType({
             const service = accessory.addService(Service.AirQualitySensor, device.name);
```
koush, line 104: needs to be moved into the if statement if one of those interfaces exists.

koush: disregard, had to expand diff to see the whole file. I see it now.

PR #259, opened 2022-05-23
homekit: add co2sensor

billyzoellers:
Adds CO2 sensor, with "CO2 detected" when PPM is greater than 5000


plugins/homekit/src/types/sensor.ts
```diff
@@ -115,6 +115,15 @@ addSupportedType({
             }
         }
 
+
+        if (device.interfaces.includes(ScryptedInterface.CO2Sensor)) {
+            const service = accessory.addService(Service.CarbonDioxideSensor, device.name);
+            bindCharacteristic(device, ScryptedInterface.CO2Sensor, service, Characteristic.CarbonDioxideLevel,
+                () => device.co2ppm || 0);
+            bindCharacteristic(device, ScryptedInterface.CO2Sensor, service, Characteristic.CarbonDioxideDetected,
+                () => ((device.co2ppm || 0) > 5000))
```
billyzoellers, line 124: @koush - only thing that I think is questionable, since I'm picking a value here (5000) that causes an alert. Is there a different way you want to handle?

koush: any idea how other HomeKit accessories handle it? 5000 seems to be the google number.

PR #268, opened 2022-05-29
homekit: set security system state valid values

billyzoellers:


plugins/homekit/src/types/security.ts
```diff
@@ -16,6 +16,19 @@ addSupportedType({
         const service = accessory.addService(Service.SecuritySystem, device.name);
         service.setPrimaryService();
 
+        // Set available modes based on plugin
+        function systemStates(supportedModes: Array<SecuritySystemMode>): number[] {
```
koush, line 20: this function is unnecessary since this is only called once.

PR #292, opened 2022-06-24
Add security workflows

bryopsida:
This adds two workflows to do security scans on PRs and on a cron.

Workflows added
---
- Trivy - Scans the docker image and performs SCA on the OS packages and software manifests (package.json files) and provides findings (packages with known vulnerabilities) to the GitHub security tab
- CodeQL - Scans python and javascript code for bugs and insecure code in this repository and updates any findings in the GitHub security tab.

Why
---
I run automatic scans on things I run or am considering running and wanted to provide these workflows to increase awareness of potential issues.

Questions
---
Is this something this project wants? If so let me know what else I need to do to get this PR accepted. cc @koush


koush:
As this is (currently) a mono repo with a lot of optional plugins, I'd prefer this to run on only the core plugins and server.

koush:
> As this is (currently) a mono repo with a lot of optional plugins, I'd prefer this to run on only the core plugins and server.

Also for this reason, push trigger may not be appropriate. I'd probably change the workflow to run on npm publish of the server.

bryopsida:
> Also for this reason, push trigger may not be appropriate. I'd probably change the workflow to run on npm publish of the server.

What about dropping the on push to pr and just using a cron, maybe nightly, or weekly if nightly is too frequent? One reason to scan on a schedule is the CVE databases/and CodeQL queries change frequently so only running on publish would only capture what was known at that time which may not accurately represent the current state depending on how long the gap is between publishes.

bryopsida:
> As this is (currently) a mono repo with a lot of optional plugins, I'd prefer this to run on only the core plugins and server.

Would excluding anything under the plugins folder suffice for this?

koush:
these plugins should be included, and the "packages" folder

image

PR #311, opened 2022-07-13
Update README.md

townsmcp:
Added step by step guide to help others set up Google Device Access


koush:
This looks good, but I think having a "TLDR" version should also remain so people don't get intimidated by the 60 step process. I'll make that change before the next plugin update, if you can't get to it.

townsmcp:
@koush sure, that makes sense. On it now. Will close this pull request, make some changes and then  submit a new pull (I havent done this before so think thats what Im supposed to do)

PR #350, opened 2022-08-23
Add Tuya Camera (and Doorbell Cameras) Support

ErrorErrorError:
Hello,

I created a plugin that allows users to discover Tuya cameras registered in Tuya IoT Cloud platform and allows you to view and control cameras via Scrypted.

This plugins require that all cameras are registered in Tuya IoT Cloud and you will need to create an account with them. For more information please check out the readme that is [here](https://github.com/ErrorErrorError/scrypted/blob/main/plugins/tuya/README.md).

TODOs
- Support 2 way audio (Can only be achieved with WebRTC)
- Support doorbell presses

Please I do need some testing with other devices since I only have a `Feit Floodlight Camera` but I do want to bring support for other cameras. If you are having issues please let me know on Discord!


plugins/tuya/src/camera.ts
```diff
@@ -0,0 +1,209 @@
...
+        this.binaryState = true;
+        setTimeout(() => this.binaryState = false, 10000);
+    }
+
+    // most cameras have have motion and doorbell press events, but dont notify when the event ends.
+    // so set a timeout ourselves to reset the state.
+
+    triggerMotion() {
+        this.motionDetected = true;
+        setTimeout(() => this.motionDetected = false, 10000);
```
koush, line 171: this will have funny behavior if motion is triggered multiple times, should maybe use a debounce of some sort to keep the motion active.

plugins/tuya/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
koush, line 1: not sure if this file is actually used, the scrypted plugin build process just assumes a specific file structure and has a tsconfig.json under the hood.

plugins/tuya/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Follow this [guide](https://developer.tuya.com/en/docs/iot/Configuration_Guide_custom?id=Kamcfx6g5uyot&_source=bdc927ff355af92156074d47e00d6191)
+
+Once you have retreived both the `Access Id` and `Access Key` from the project, you can get the `User Id` by going to Tuya Cloud IoT -> Select the Project -> Devices -> Link Tuya App Account -> and then get the UID.
+
+## TODOs
+- Fix 2-way talk for supported platforms (Can only work with WebRTC since we only get one stream with RTSPS)
+- Add support for camera doorbells (Just need to implement doorbell notification)
+
+## npm commands
```
koush, line 19: can remove all this development related stuff, I've been moving that into a separate Development.md.

plugins/tuya/src/main.ts
```diff
@@ -0,0 +1,284 @@
...
+            const ret = new TuyaCamera(this, nativeId, camera);
+            this.cameras.set(nativeId, ret);
+            return ret;
+        }
+
+        throw new Error('device not found?');
+    }
+}
+
+export default createInstanceableProviderPlugin("Tuya", nativeId => new TuyaController(nativeId));
```
koush, line 284: I'm not sure this is necessary unless you want to explicitly support multiple logins? you may have just copied this from unifi?

plugins/tuya/src/camera.ts
koush: I think this was a bug I had in my camera plugins if that's where you got the code, but it was recently fixed in most of them

ErrorErrorError: Oops, I actually had that in mind but forgot to add it

plugins/tuya/tsconfig.json
ErrorErrorError: cool, Ill remove it

plugins/tuya/src/camera.ts
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    // VideoCamera
+
+    async getVideoStream(
+        options?: MediaStreamOptions
+    ): Promise<MediaObject> {
+        const vsos = await this.getVideoStreamOptions();
+        const vso = vsos.find(find => find.id === options.id) || vsos[0];
```
koush, line 106: no need for this search, your video stream options will always be the same, since there's only 1 stream. unifi provides 3 streams.

plugins/tuya/src/main.ts
ErrorErrorError: In Tuya IoT Cloud, you can create multiple projects and each of those projects can contain multiple devices, however not sure if anyone would really do that

plugins/tuya/README.md
ErrorErrorError: will do!

plugins/tuya/src/main.ts
koush: k up to you, it's common in large unifi deployments.

ErrorErrorError: I'll keep it in the mean time

ErrorErrorError:
Please review as much as you need! I fixed the ones you pointed out and cleaned up unused imports

fennec622:
Hooooooooo

Very good news !!!! I have Two doorbell, if I can use two way audio with HomeKit amazing

You have a doorbell HVS for 50$ !!!!

Faster I test and I feed back

ErrorErrorError:
> Hooooooooo
> 
> 
> 
> Very good news !!!! I have Two doorbell, if I can use two way audio with HomeKit amazing
> 
> 
> 
> You have a doorbell HVS for 50$ !!!!
> 
> 
> 
> Faster I test and I feed back 
> 
> 
> 
> i have two 
> 
> LSC-Bell 8S(Mstar333_8M)

Two way audio not supported at the moment, but I will implement it soon, and you might get an error trying to add devices. There will be an update that fixes that issue and that will support two way audio

rxvincent:
The Tuya IOT platform has only 1 month for trial?

PR #393, opened 2022-10-03
added swing state/status to the Fan, to support HK SwingMode

nberardi:


nberardi:
Added the fan swing state/status in order to support HomeKit SwingMode

koush:
What is swing mode?

nberardi:
[SwingMode](https://developer.apple.com/documentation/homekit/hmcharacteristictypeswingmode) is the ability of a table top fan to move back and forth from left to right.

PR #395, opened 2022-10-05
added support for greater capabilities in HomeKit for Fan's

nberardi:


koush:
Seems fine, but the code from the sensors shouldn't be duplicated into the thermostat/fan etc. Can you consolidate that into  a common file that can be used by whatever ScryptedDeviceType needs it? Ie, Thermostat or Sensor.

nberardi:
That was a good idea, I have incorporated those changes into the latest update to this pull request

koush:
one more change I think you forgot: can you rewire the existing Sensor type to use the new common code? Will get coverage that way as I have sensors that I can test with.

nberardi:
There is a possibility I misunderstood you. I did move the sensors to a common extension method for both fan and thermostat.

koush:
looks like air quality and co/co2 sensors

koush:
I'll take care of it in the morning if you don't get to it, it's a straightforward change that I may not be explaining properly. its outside of the code you used/changed, and am just now noticing it was duplicated.

PR #437, opened 2022-11-23
feat: allow Arlo basestations that identify as sirens (e.g. VMB4000)

brianschrameck:
The Arlo plugin attempts to discover the associated basestation(s) for the camera(s). Older basestations (like the VMB4000) advertise themselves as `siren` instead of `basestation`, so this PR allows these basestations to be discovered correctly. Tested working with Arlo Pro 2 cameras paired to VMB4000v3.

As a side note, I don't get any audio, but not sure if that's a plugin problem or a problem with how the basestation is advertising itself.


koush:
@bjia56

bjia56:
Unfortunately I don't have the same model to be able to confirm, but this change looks reasonable to me. I can publish once it gets merged in.

Issue #445, opened 2022-11-29
use relative url for video clip download links


lvnilesh:
The dashboard opens properly https://scrypted.local.domain.com/ with a proper certificate and all but the video clips download button shows https://192.168.1.4:10443/endpoint/8/public/guid.mp4 and fail with a wrong cert

koush:
duplicate

koush:
f3f370d5

PR #447, opened 2022-11-30
Add SCRYPTED_HOSTNAME setting

### LINKED ISSUE #367: Video clips don't display when accessing Scrypted via hostname
I noticed in the source code that Scrypted is rendering html as:
```
<video src="https://192.168.1.1:1234/somepath/clip.mp4" controls="controls" autoplay="autoplay"></video>
```

But that means videos don't display when I'm accessing Scrypted via `https://hostname.local:1234`. I'm hoping it's trivial for the app to switch to root-relative urls instead like:

```
<video src="/somepath/clip.mp4" controls="controls" autoplay="autoplay"></video>
```

jaysoffian:
When using Scrypted with https and a hostname, video clips fail to display because the URL that Scrypted constructs uses an IP address, which is typically not included in the TLS certificate CN/Subject Alternative Name.

Add support for SCRYPTED_HOSTNAME setting which, if set, is used to construct video clip URLs instead of the server's IP address.

Fixes #367, #445.

### DIFF
--- a/server/src/plugin/plugin-remote.ts
+++ b/server/src/plugin/plugin-remote.ts
@@ -72,9 +72,13 @@ class EndpointManagerImpl implements EndpointManager {
         return id;
     }
 
-    async getUrlSafeIp() {
+    async getUrlSafeHost() {
+        const hostname: string = await this.api.getComponent('SCRYPTED_HOSTNAME');
+        if (hostname) {
+            return hostname;
+        }
         // ipv6 addresses have colons and need to be bracketed for url safety
-        const ip: string = await this.api.getComponent('SCRYPTED_IP_ADDRESS')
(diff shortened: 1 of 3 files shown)


jaysoffian:
@koush I'm really not sure if this is the best way to do this. I'm not clear what else the various `getFooEndpoint` methods are used for. It would also be great if a new setting wasn't needed, so perhaps the code could defer to the HTTP Host header or construct a relative instead of absolute URL. But since I'm not very familiar with the code base, I took a very conservative approach here where the existing behavior does not change at all unless `SCRYPTED_HOSTNAME` is set via the environment.

koush:
Correct solution here is to use relative paths in the browser context: f3f370d5

PR #463, opened 2022-12-12
First working version for a BTicino C300X

slyoldfox:


koush:
This is duplicating al to of the code from the sip plugin, many of the files are straight up copied. Please use relative imports to source files within the sip plugin. Ie:

import * from '../../sip/src/rtp-utils';

It will be easier to examine this as a diff, and also share fixes across both plugins.

PR #468, opened 2022-12-14
First working version for a BTicino C300X

slyoldfox:
Follow up of #463 - might extract some more code from sip-call an/or sip-session as I move on to audio out

### DIFF
--- a/plugins/bticino/src/main.ts
+++ b/plugins/bticino/src/main.ts
@@ -42,81 +42,56 @@ class SipCamera extends ScryptedDeviceBase implements Intercom, Camera, VideoCam
         return;
     }
 
-    storageSettings = new StorageSettings(this, {})
+    settingsStorage = new StorageSettings(this, {
+        sipfrom: {
+            title: 'SIP From: URI',
+            type: 'string',
+            value: this.storage.getItem('sipfrom'),
+            description: 'SIP URI From field: Using the IP address of your server you will be calling from. Also the user and IP you added in /etc/flexisip/users/route_ext.conf on the intercom.',
+            placeholder: 'user@192.168.0.111',
+            multiple: false,
+        },
+        sipto: {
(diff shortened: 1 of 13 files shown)


koush:
Thanks, much better. Why isn’t sip call and sip session from ring being used as well? Were there changes?

slyoldfox:
Well. That is all about the bticino requiring a SIP REGISTER (I implemented this). But there is also a whole bunch of hacks I had to add to work around non-implemented code and bugs in sip.js (which at this point I don't want to modify there seeing how old that code is).

When this goes through I'll start on the audio out part - i'll probably extract that part from sip plugin in a utils class and see how far I can refactor sip-call and sip-session.

If you need more info about the hacks I can discuss them with you.

plugins/bticino/src/main.ts
```diff
@@ -0,0 +1,381 @@
...
+    getSipMediaStreamOptions(): ResponseMediaStreamOptions {
+        return {
+            id: 'sip',
+            name: 'SIP',
+            // this stream is NOT scrypted blessed due to wackiness in the h264 stream.
+            // tool: "scrypted",
+            container: 'sdp',
+            video: {
+                codec: 'h264',
+                h264Info: {
```
koush, line 291: is this accurate or did you copy this from somewhere? May be better to leave it blank and let rebroadcast handle it.

slyoldfox: It was a shameless copy from "sip". I tried just remove the h264Info part but that gave me following errors:

```
[BTicino SIP Plugin] unknown nal unit type 31

[BTicino SIP Plugin] unknown nal unit type 0

[BTicino SIP Plugin] unknown nal unit type 27

[BTicino SIP Plugin] unknown nal unit type 25

[BTicino SIP Plugin] unknown nal unit type 27
```
So I opted to remove it completely which works.

PR #479, opened 2022-12-27
added support for type `Garage` and refactored the controller for future support

nberardi:
I took a stab and updating the Alexa plugin, and start setting it up for expanded device support.  There was also a few things that could have been handled better around authentication and messageId that better aligned with the developer documentation that I implemented as well.  I have verified that existing support for `Camera` and `Doorbell` has continued to work.  In addition I have added support for `Garage`.

### DIFF
--- a/plugins/alexa/src/main.ts
+++ b/plugins/alexa/src/main.ts
@@ -4,7 +4,7 @@ import { StorageSettings } from '@scrypted/sdk/storage-settings';
 import { AutoenableMixinProvider } from '@scrypted/common/src/autoenable-mixin-provider';
 import { isSupported } from './types';
 import { DiscoveryEndpoint, DiscoverEvent } from 'alexa-smarthome-ts';
-import { AlexaHandler, capabilityHandlers, supportedTypes } from './types/common';
+import { AlexaHandler, addBattery, addOnline, addPowerSensor, capabilityHandlers, supportedTypes } from './types/common';
 import { createMessageId } from './message';
 
 const { systemManager, deviceManager } = sdk;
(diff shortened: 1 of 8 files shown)


plugins/alexa/src/main.ts
```diff
@@ -190,18 +214,41 @@ class AlexaPlugin extends AutoenableMixinProvider implements HttpRequestHandler,
...
+                        self.log.w(error?.response?.data?.error_description);
+                        break;
+                    
+                    case 'expired_token':
+                        self.accessToken = undefined;
+                        break;
+
+                    default:
+                        self.console.error(error?.response?.data);
+                }
```
koush, line 243: modified this change to rethrow the error here. otherwise it returns an undefined response which then fails obscurely.

koush:
I did start seeing errors that my token needed to be refreshed, though Alexa continued working even though the plugin was complaining. I'm guessing that the state reports had been failing silently. I"m not really an Alexa user, and mostly just wrote this plugin as an exercise for webrtc compatibility testing on the cameras.

plugins/alexa/src/main.ts
```diff
@@ -214,19 +261,47 @@ class AlexaPlugin extends AutoenableMixinProvider implements HttpRequestHandler,
...
-                "header": {
-                    "namespace": "Alexa.Authorization",
-                    "name": "AcceptGrant.Response",
-                    "messageId": createMessageId(),
-                    "payloadVersion": "3"
-                },
-                "payload": {}
+        if (accessToken !== undefined) {
+            try {
+                response.send(JSON.stringify({
```
koush, line 288: I don't think it's possible for this block to fail.

plugins/alexa/src/main.ts
```diff
@@ -214,19 +261,47 @@ class AlexaPlugin extends AutoenableMixinProvider implements HttpRequestHandler,
...
-        response.send(JSON.stringify({
-            "event": {
-                "header": {
-                    "namespace": "Alexa.Authorization",
-                    "name": "AcceptGrant.Response",
-                    "messageId": createMessageId(),
-                    "payloadVersion": "3"
-                },
-                "payload": {}
+        if (accessToken !== undefined) {
```
koush, line 286: access token will always be defined, since failure will throw now (and be handled as a 500 down below)

plugins/alexa/src/main.ts
nberardi: I was going to let these get swallowed, as they are non-actionable by us at the moment.

plugins/alexa/src/main.ts
nberardi: This was failing during the debug process if the response didn’t come before the timeout.  Theoretically this is a debug problem, however it could occur in the wild.

PR #498, opened 2023-01-06
Tweak node check to >=18

longzheng:
Otherwise will fail on Node 19


koush:
This is intentional and prevents known checkout issues (join #development on Discord). Node 19 is not supported due to missing prebuilt native dependencies and has caused issues with the script itself. Node 18 is LTS. Node 19 (odd number) will not be LTS. Node 20 will be LTS. Will update when that releases.

longzheng:
Ah right. When I ran the Windows script it auto installed Node 19 for me (I was on Node 18). I'll open a new PR to update that to install specifically Node 18.

PR #502, opened 2023-01-07
Windows installation script install Node LTS v18

longzheng:
Otherwise `OpenJS.NodeJS` defaults to v19 right now which is not supported https://github.com/koush/scrypted/pull/498


koush:
In retrospect, would be better to do the specific version, 18, if possible. Can add that when you include the command to install the node build tools as well.

longzheng:
Sure thing.

PR #503, opened 2023-01-07
Update Windows installation script

longzheng:
- Changes to pinned version of Node LTS 18.13.0 https://github.com/koush/scrypted/pull/502#issuecomment-1374325481
- Change to use Chocolatey to install Node and Python (removes dependency on `winget` since the Node.js additional tools already use Chocolatey)
- Remove `windows-build-tools` and install the Node.js additional tools for Windows to compile native modules as recommended by https://github.com/felixrieseberg/windows-build-tools using the official script https://github.com/nodejs/node/blob/main/tools/msvs/install_tools/install_tools.bat#L55 
- Simplify service installer Node script
- Fix service start not working and replace with `sc start`

We can also simplify the uninstallation instructions at https://github.com/koush/scrypted/wiki/Uninstallation#windows to use [`sc.exe`](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/sc-delete) directly which avoids the dependency on `node-windows`
```cmd
sc stop scrypted.exe
sc delete scrypted.exe
```

### DIFF
--- a/docker/install-scrypted-dependencies-win.ps1
+++ b/docker/install-scrypted-dependencies-win.ps1
@@ -1,5 +1,16 @@
-winget install -h --id "OpenJS.NodeJS.LTS"
-winget install -h --id "Python.Python.3.10"
+# Install Chocolatey
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
+
+# Install node.js
+choco upgrade -y nodejs-lts --version=18.13.0
+
+# Install Node.js additional tools for Windows to compile native modules
+# https://github.com/nodejs/node/blob/main/tools/msvs/install_tools/install_tools.bat#L55
+choco upgrade -y python visualstudio2019-workload-vctools
+
+# Refresh environment variables for py and npx to work
+$env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User") 
 
 py -m pip install --upgrade pip
 py -m pip install aiofiles debugpy typing_extensions typing opencv-python
@@ -9,7 +20,6 @@ npx -y scrypted@latest install-server
 $USER_HOME_ESCAPED = $env:USERPROFILE.replace('\', '\\')
 $SCRYPTED_HOME = $env:USERPROFILE + '\.scrypted'
 $SCRYPTED_HOME_ESCAPED_PATH = $SCRYPTED_HOME.replace('\', '\\')
-npm install --global --production windows-build-tools
 npm install --prefix $SCRYPTED_HOME node-windows@1.0.0-beta.8 --save
 
 $SERVICE_JS = @"
@@ -19,7 +29,6 @@ try {
 }
 catch (e) {
 }
-
 const child_process = require('child_process');
 child_process.spawn('npx.cmd', ['-y', 'scrypted', 'serve'], {
     stdio: 'inherit',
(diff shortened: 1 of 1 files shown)


docker/install-scrypted-dependencies-win.ps1
```diff
@@ -44,30 +52,19 @@ var svc = new Service({
...
-svc.on('error', () => {
-  svc.install();
-});
-
 svc.on('install', () => {
-  svc.start();
+  console.log("Service installed");
 });
-
-svc.uninstall();
```
longzheng, line 64: I found this quite redundant so simplified, not sure if there was a reason for all the other event listeners.

docker/install-scrypted-dependencies-win.ps1
```diff
@@ -44,30 +52,19 @@ var svc = new Service({
...
+svc.install();
 "@
 
 $INSTALL_SERVICE_JS_PATH = $SCRYPTED_HOME + '\install-service.js'
 $INSTALL_SERVICE_JS | Out-File -Encoding ASCII -FilePath $INSTALL_SERVICE_JS_PATH
 
 node $INSTALL_SERVICE_JS_PATH
 
+# Manually start service, node-windows has issues starting service
+sc start scrypted.exe
```
longzheng, line 67: The `svc.start()` script didn't work for me. Even though `node-windows` reports is started successfully, I found that the service showed as "stopped" in the service application. I find `sc start` work more reliably.

docker/install-scrypted-dependencies-win.ps1
koush: This code would trigger an install after uninstalling it, or attempting, first. The intention is to make sure the service script can update itself.

koush:
I’m not familiar with chocolates, but seems fine as long as that dependency is soft installed.

longzheng:
> I’m not familiar with chocolates, but seems fine as long as that dependency is soft installed.

Yeah chocolatey and winget are very similar, chocolatey has a bit more mature.

docker/install-scrypted-dependencies-win.ps1
longzheng: Ah ok. There seems to be a bug or something with `node-windows` uninstall at the moment that causes an exception https://github.com/coreybutler/node-windows/issues/324 even though the service is successfully uninstalled.

longzheng: Ah figured out a workaround.

docker/install-scrypted-dependencies-win.ps1
longzheng: I ended up removing this since I fixed `node-windows` itself

koush:
Change ready to go then?

longzheng:
Yep I've tested the latest script using a new Windows Sandbox and it's one click to get the server running now 😇

koush:
I updated the uninstallation instructions. It doesn't need a full path to the scrypted.exe?

longzheng:
> I updated the uninstallation instructions. It doesn't need a full path to the scrypted.exe?

No because `scrypted.exe` is actually the ID name of the generated service by `node-windows`. 

Even though we specify the `name` parameter, that's actually the friendly name which is not used by `sc.exe`

```

SERVICE_NAME: scrypted.exe
DISPLAY_NAME: Scrypted
        TYPE               : 10  WIN32_OWN_PROCESS
        STATE              : 4  RUNNING
                                (STOPPABLE, NOT_PAUSABLE, ACCEPTS_SHUTDOWN)
        WIN32_EXIT_CODE    : 0  (0x0)
        SERVICE_EXIT_CODE  : 0  (0x0)
        CHECKPOINT         : 0x0
        WAIT_HINT          : 0x0
```

PR #530, opened 2023-01-23
CoreML: Get input tensor size automatically

Roarrk:
Get/Save MLMODEL spec automatically. This can also later be used for getting labels automatically.


plugins/coreml/src/coreml/__init__.py
```diff
@@ -35,11 +40,11 @@ def __init__(self, nativeId: str | None = None):
     # width, height, channels
     def get_input_details(self) -> Tuple[int, int, int]:
         # TODO: is there a way to get this from the model itself?
```
koush, line 42: can remove these comments as they are not defunct

PR #545, opened 2023-02-03
rebroadcast: handle the case where server_port from camera is 0

brianschrameck:
When connecting directly to Arlo cameras (maybe others?), the opus track returns `server_port=0-0`. This small change allows the Scrypted RTSP parser to skip these tracks and allows the camera to connect successfully.


koush:
Should ideally figure out what is incorrect about the scrypted rtsp negotiation.

8cc8fbfd547724fe01b6ca2a99a3f70da48ab597

PR #547, opened 2023-02-08
Accomodate custom models

Roarrk:
Tries to fetch the label list from the mlmodel file itself and due to newer converted mlmodels having a different output (float32) it also numpy converts the model output to floats.

### DIFF
--- a/plugins/coreml/src/coreml/__init__.py
+++ b/plugins/coreml/src/coreml/__init__.py
@@ -18,6 +18,9 @@ def parse_label_contents(contents: str):
             ret[row_number] = content.strip()
     return ret
 
+def list_to_dict(list1):
+    return {i: list1[i] for i in range(len(list1))}
+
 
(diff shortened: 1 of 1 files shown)


plugins/coreml/src/coreml/__init__.py
```diff
@@ -47,11 +56,11 @@ def get_input_size(self) -> Tuple[float, float]:
     def detect_once(self, input: Image.Image, settings: Any, src_size, cvss):
         out_dict = self.model.predict({'image': input, 'confidenceThreshold': .2 })
 
-        coordinatesList = out_dict['coordinates']
+        coordinatesList = out_dict['coordinates'].astype(float)
```
koush, line 59: do we need to type check this first? dont previous coremls return a proper float array?

koush:
Reverted this change @Roarrk
the labels fail to load properly in the included model.

PR #553, opened 2023-02-12
New DahuaEvent and Introduction of CallerId for Dahua Intercoms

KoljaV:
- New DahuaEvent to reset Binary State of Bell, when Call is denied
- Introduced CallerId to check the Calling Event for this CallerId. Some Dahua Intercoms support multiple Bell Buttons and each of these Bell Buttons call a different CallerId.


koush:
Looks like this setting is used even if multiple call ids is disabled. Also typo in description.

KoljaV:
@koush You are right! Let me correct that one.

koush:
any update on this @KoljaV?

PR #554, opened 2023-02-12
Add JSON output for webhook methods calls

kiler129:


plugins/webhook/src/main.ts
```diff
@@ -112,7 +112,16 @@ class WebhookMixin extends SettingsMixinDeviceBase<Settings> {
                 }
                 
                 const result = await device[methodOrProperty](...parameters);
-                this.maybeSendMediaObject(response, result, methodOrProperty);
+                if (request.headers['accept'] && request.headers['accept'].indexOf('application/json') !== -1) {
```
koush, line 115: Could optional chain ?.includes

plugins/webhook/src/main.ts
```diff
@@ -92,7 +92,7 @@ class WebhookMixin extends SettingsMixinDeviceBase<Settings> {
             }
         });
     }
-
+
```
koush, line 95: Whitespace

kiler129:
@koush Fixed ;) I also changed the other instance to `?.includes()` which is indeed more elegant.

PR #559, opened 2023-02-14
prebuffer: allow battery-powered cameras to prebuffer if they have external power applied

brianschrameck:
This is a small change to allow cameras to implement the `Battery` interface, but still be able to prebuffer if they are plugged in. This allows battery levels to continue to be reported to Scrypted when plugged in, which is useful while charging battery-powered cameras. It also prevents having to change the interfaces that a device implements when a power source changes, which causes problems when switching power sources and the camera is streaming.

### DIFF
--- a/plugins/prebuffer-mixin/src/main.ts
+++ b/plugins/prebuffer-mixin/src/main.ts
@@ -1469,7 +1469,7 @@ class PrebufferMixin extends SettingsMixinDeviceBase<VideoCamera> implements Vid
       }
     }
 
-    const isBatteryPowered = this.mixinDeviceInterfaces.includes(ScryptedInterface.Battery) && !this?.externallyPowered;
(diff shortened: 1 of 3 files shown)


plugins/prebuffer-mixin/src/main.ts
```diff
@@ -1469,7 +1469,7 @@ class PrebufferMixin extends SettingsMixinDeviceBase<VideoCamera> implements Vid
       }
     }
 
-    const isBatteryPowered = this.mixinDeviceInterfaces.includes(ScryptedInterface.Battery);
+    const isBatteryPowered = this.mixinDeviceInterfaces.includes(ScryptedInterface.Battery) && !this?.externallyPowered;
```
brianschrameck, line 1472: Not sure how this optional got in here.

koush:
power/charging should be a separate interface. in addition, this is not accurate, as rings are can be externally powered and still drain battery faster (and die) while prebuffering.

I think adding a property to ResponseVideoStream or whatever it's called is the best approach here. something like forcePrebuffer.

PR #573, opened 2023-02-22
fix ring alarm state when changing modes

leedsalex:
I have no idea what is being done in homebridge, but this solution I came up with definitely works for me.
I did find that the `location.onLocationMode.subscribe` is explicitly **not** used if the location has a base station, which my location does. It's possible we might be over updating the location mode with this change for some users if they don't have a base station and also have a security panel, but it wasn't updating at all previously for my very simple ring setup. I'm also not sure what type of ring location doesn't have a base station but still can be armed, that part is a little confusing.

### DIFF
--- a/plugins/ring/src/main.ts
+++ b/plugins/ring/src/main.ts
@@ -699,8 +699,9 @@ export class RingLocationDevice extends ScryptedDeviceBase implements DeviceProv
                     updateLocationMode(response.mode);
                 });
             });
-        }).catch(_ => {
-            this.console.warn(`no security panel found, updates will not be recieved on this subscription`)
+        }).catch(error => {
+            // could not find a security panel for location
+            // not logging this error as it is a valid case to not have a security panel
         });
 
         if (location.hasAlarmBaseStation) {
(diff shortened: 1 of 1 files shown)


plugins/ring/src/main.ts
```diff
@@ -690,6 +690,18 @@ export class RingLocationDevice extends ScryptedDeviceBase implements DeviceProv
...
+        // if the location has a base station, updates when arming/disarming are not sent to the `onLocationMode` subscription
+        // instead we subscribe to the security panel, which is updated during arming actions
+        location.getSecurityPanel().then(panel => {
+            panel.onData.subscribe(_ => { 
+                location.getLocationMode().then(response => {
+                    updateLocationMode(response.mode);
+                });
+            });
+        }).catch(_ => {
+            this.console.warn(`no security panel found, updates will not be recieved on this subscription`)
```
koush, line 703: is there a hasSecurityPanel you can check before invoking this? users will get weirded out by debug messages that look like errors.

leedsalex: There is not, I will just not log the error and put a comment in the catch statement

PR #576, opened 2023-02-23
alexa plugin: ensure we are talking to the right endpoint region for a customer

nberardi:
**DO NOT MERGE**

I still need to do some targeted testing within the NA, EU, and FE regions.  However I am posting this for early access review to get more eyes on the change.


koush:
Whoops, I should have read first fully first. immediately went into the diff to review, and then merged. You can use draft pull requests if you do not intend to merge.

Issue #482, opened 2022-12-29
Feature request: Allow Ring Alarm arming to automatically bypass sensors


craftbyte:
When trying to arm my Ring Alarm, I sometimes leave the window of my bedroom slightly open, as I sleep in the room anyway. On the Ring app and Keypad, I can manually bypass the sensor after trying to arm. Can you make an option in settings that will automatically bypass the sensors that are needed to be bypassed for arming the system when arming from HomeKit?
This is already supported in HomeBridge by adding a Night mode to the toggle, that is an automatic Bypass and Arm.

PR #577, opened 2023-02-23
ring: added night mode support to arm with sensor bypass

leedsalex:
closes https://github.com/koush/scrypted/issues/482

This PR also fixes two unrelated warnings I noticed in the homekit logs for the garage and window covering type. The following was not needed as we already call `bindCharacteristic`
```
service.getCharacteristic(Characteristic.TargetDoorState)
.on(CharacteristicEventTypes.GET, (callback: NodeCallback<CharacteristicValue>) => {
    callback(null, targetState);
})
```

### DIFF
--- a/plugins/homekit/src/types/garage.ts
+++ b/plugins/homekit/src/types/garage.ts
@@ -22,9 +22,6 @@ addSupportedType({
 
         let targetState = !!device.entryOpen ? Characteristic.TargetDoorState.OPEN : Characteristic.TargetDoorState.CLOSED;
         service.getCharacteristic(Characteristic.TargetDoorState)
-            .on(CharacteristicEventTypes.GET, (callback: NodeCallback<CharacteristicValue>) => {
-                callback(null, targetState);
-            })
             .on(CharacteristicEventTypes.SET, (value: CharacteristicValue, callback: CharacteristicSetCallback) => {
                 callback();
                 if (value === Characteristic.TargetDoorState.OPEN) {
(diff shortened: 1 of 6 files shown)


plugins/homekit/src/types/garage.ts
```diff
@@ -22,9 +22,6 @@ addSupportedType({
 
         let targetState = !!device.entryOpen ? Characteristic.TargetDoorState.OPEN : Characteristic.TargetDoorState.CLOSED;
         service.getCharacteristic(Characteristic.TargetDoorState)
-            .on(CharacteristicEventTypes.GET, (callback: NodeCallback<CharacteristicValue>) => {
```
koush, line 25: is this deletion intentional? not sure why garage is being changed.

plugins/homekit/src/types/windowcovering.ts
```diff
@@ -28,9 +28,6 @@ addSupportedType({
         let targetState = !!device.entryOpen ? 100 : 0;
         service.getCharacteristic(Characteristic.TargetPosition)
             .setProps(props)
-            .on(CharacteristicEventTypes.GET, (callback: NodeCallback<CharacteristicValue>) => {
```
koush, line 31: what's this?

koush:
ring changes look fine, but I dont know what some of those homekit changes are (one looks like a brevity cleanup)

plugins/homekit/src/types/garage.ts
leedsalex: I made a comment in the pr description. Those are unneeded, as we are calling the callback twice per event. Once from bind characteristic and then again here. But the bind characteristic handling is much better. The double callback is causing warning all over the logs.

plugins/homekit/src/types/windowcovering.ts
leedsalex: Same as above

PR #583, opened 2023-02-25
add fan support to nest thermostats

leedsalex:


plugins/google-device-access/src/main.ts
```diff
@@ -425,6 +432,32 @@ class NestThermostat extends ScryptedDeviceBase implements HumiditySensor, Therm
         // not supported by API. throw?
     }
 
+    async turnOff(): Promise<void> {
+        // You can't turn the fan off when the HVAC unit is currently running.
+        if (this.thermostatActiveMode !== ThermostatMode.Off) {
+            await this.refresh(null, true); // Refresh the state to turn the fan switch back to active.
```
leedsalex, line 438: I could use some help on this, basically if the user "turns off" the fan switch in homekit, and the AC is currently running, I want the switch to turn back on "immediately" in the homekit UI

koush: I think you'd need to set it `this.on=false` and then let the refresh turn it back on.

leedsalex: I think `this.on` is already in the `false` state at the point this method is hit. It might just be some sort of lag

leedsalex: Well that worked :)

koush:
@leedsalex gave you publish access on npm

PR #587, opened 2023-02-26
New DahuaEventType and introduction of CallerId for Dahua Intercoms with multiple Bell Buttons. 

KoljaV:


KoljaV:
Found an additional issue. Need to fix it. Pls do not merge.

PR #617, opened 2023-03-10
eufy: allow setting livestream duration

leedsalex:


koush:
why is this necessary? I think the stream should remain open until the rtsp server session ends.

koush:
fork or set to infinitely, this value shouldn't be configurable. the stream should end when the rtsp client (rebroadcast in this case) stops streaming.

koush:
also I noticed the battery interface was on this camera, which was interfering with the camera pre buffer starting. is this intentional? is it not wired?

PR #639, opened 2023-03-18
hap: fix camera init without sensor

leedsalex:


koush:
What’s with the myQ change

leedsalex:
oh woops, no idea lol... but isn't that in another org anyway?

leedsalex:
fixed

PR #650, opened 2023-03-22
homekit: merge SecuritySystem & child Sirens as one Homekit accessory

bjia56:
This exposes the Scrypted OnOff interface to Homekit's SecuritySystem as a Switch. The Homekit UI will now show both the Arm/Disarm selector as well as the switch.

The intended use is for Arlo's basestation siren to emulate an arm/disarm to reduce the risk of being accidentally toggled.

### DIFF
--- a/plugins/homekit/src/main.ts
+++ b/plugins/homekit/src/main.ts
@@ -184,12 +184,17 @@ export class HomeKitPlugin extends ScryptedDeviceBase implements MixinProvider, 
         }
 
         const plugins = await systemManager.getComponent('plugins');
-
         const accessoryIds = new Set<string>();
-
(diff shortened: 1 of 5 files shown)


koush:
I hesitant to do it this way, having implemented a similar bit with cameras with lights on them (and I'd like to undo this). For example, I now have a camera with a siren AND a light. Which one gets the OnOff?

The way I think it should work:

The security system is a DeviceProvider with a child siren device.
The HomeKit plugin sees this siren on a security system and merges that into the accessory as a switch.

I think devices like security system and sirens should be discrete in Scrypted, but DeviceProviders and their children merged in their UI representation in HomeKit plugin.

PR #653, opened 2023-03-24
actions: tests for platform-specific install scripts

bjia56:
I'm open to discussing whether or not something like this would be useful to have in general, but I figured it could be helpful to kick off a set of tests to verify scrypted installs on different platforms.

Right now this only tests the installation, but I intend to add more functional testing, including installing plugins that need system dependencies.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,24 @@
+name: Test
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  test_linux_local:
+    name: Test Linux local installation
+    runs-on: ubuntu-latest
+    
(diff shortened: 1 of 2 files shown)


bjia56:
This might end up being a noisy check given its current triggers, but I can clean it up a bit to run less on most changes.

koush:
wild, does the stay running on the GitHub action server?

koush:
should probably have this be manually triggered (or triggered if files in docker folders change), since I infrequently change the install scripts.

bjia56:
> wild, does the stay running on the GitHub action server?

Github actions will spin up VMs for each job so I believe once the job completes they are deleted. The job has a max lifetime of 6 hours before it's killed, so we can add and run more tests on the scrypted server as part of the actions

PR #684, opened 2023-04-02
unifi: added native zoom capability

nberardi:


nberardi:
This is ready for merge.

plugins/unifi-protect/src/camera.ts
```diff
@@ -411,6 +420,12 @@ export class UnifiCamera extends ScryptedDeviceBase implements Notifier, Interco
         if (!camera)
             return;
         this.on = !!camera.ledSettings?.isEnabled;
+        this.online = !!camera.isConnected;
         this.setMotionDetected(!!camera.isMotionDetected);
+        this.ptzCapabilities = {
```
koush, line 425: I wouldn't set state on devices that dont implement it. server doesn't reject it nor does it send out events for it, but it will be visible/persisted if checked. scrypted allows it (currently), and maybe it shouldn't. but I think it's better to not do it in unnecessarily the first place.

nberardi: That was a miss. I will correct

PR #752, opened 2023-04-22
dummy-device: mock device for Battery and Charger

bjia56:
This was useful while testing the battery prebuffer changes, to build an aggregate device with a camera + dummy battery/charger. Open to discussing if it makes sense as a separate plugin or merged into dummy-switch somehow.


koush:
I think this sort of thing would be great to include in the core plugin sample scripts.

bjia56:
Will translate this over to a sample script

PR #771, opened 2023-05-06
Inverse Bool value  Onvif motion

fennec622:
Add possibility to inverse bool value on Onvif motion


koush:
you can use a virtual device script to do this. this is not something that should be in the plugin.

PR #902, opened 2023-06-21
snapshot: tell ffmpeg pipe input format

bjia56:
Snapshots of raw images (e.g. via the snapshot url) frequently results in error when rescaling. A sample stacktrace:
```
[Snapshot Plugin]: -hide_banner -y -i pipe:[REDACTED] -filter_complex scale='min(512,iw)':-2 -frames:v 1 -f image2 pipe:3
[Snapshot Plugin]: pipe:4: Invalid data found when processing input
[Snapshot Plugin]: ffmpeg exited
[Snapshot Plugin]: Snapshot failed Error: ffmpeg input to image conversion failed with exit code: 1, /usr/bin/ffmpeg -hide_banner -y -i pipe:4 -filter_complex scale='min(512,iw)':-2 -frames:v 1 -f image2 pipe:3
[Snapshot Plugin]:     at /src/ffmpeg-image-filter.ts:177:23
[Snapshot Plugin]:     at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

It looks like this might be a side effect due to switching from static ffmpeg to system ffmpeg (maybe because static ffmpeg is a later version), since static ffmpeg doesn't produce the same error. Explicitly telling system ffmpeg to use `jpeg_pipe` makes it much happier.

### DIFF
--- a/plugins/snapshot/src/ffmpeg-image-filter.ts
+++ b/plugins/snapshot/src/ffmpeg-image-filter.ts
@@ -89,7 +89,7 @@ function ffmpegCreateOutputArguments(inputArguments: string[], options: FFmpegIm
 
 export function ffmpegFilterImageBuffer(inputJpeg: Buffer, options: FFmpegImageFilterOptions) {
     const inputArguments = [
-        '-f', 'jpeg_pipe',
+        '-f', 'image2pipe',
         '-i', 'pipe:4',
     ];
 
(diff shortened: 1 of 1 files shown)


koush:
I’m not familiar with jpeg pipe and whether all ffmpeg has that available. Typically the -f format I use for jpeg is image2. Does that work?

bjia56:
image2 doesn't work for the input arguments as it expects a list of files instead of a pipe

bjia56:
I see there's an image2pipe option, will test that out later today

koush:
Quick read indicates that the image2* demuxer/muxer handles both jpeg and png, so that's probably preferred.

PR #931, opened 2023-07-05
Improve python generation

raman325:
Adds stub references to typescript objects which are otherwise undefined, adds some formatting tweaks, and adds docstrings/comments


koush:
I think that new python code stubs should go into other.py

koush:
You seem to have checked in a bunch of package-lock.json files, can you revert those?

PR #953, opened 2023-07-14
update SystemManager.listen and listenDevice logic

raman325:
is it as simple as making this change (calling listen and listenDevice from PluginAPI)?

- Also added stubs for PluginAPI and related interfaces
- Fixed some types


koush:
`EventListenerRegister` will need to be 

```py
class EventListenerRegister:
    """Returned when an event listener is attached to an EventEmitter. Call removeListener to unregister from events."""

    def removeListener(self) -> None:
        pass
class EventListenerRegister:
    """Returned when an event listener is attached to an EventEmitter. Call removeListener to unregister from events."""

    async def removeListener(self) -> None:
        pass
```

Or you'll need to shim it on return to provide a non-async version that calls run_coroutine_threasafe on the plugin event loop.

koush:
You'll also need to do similar for the SystemManager.listen and listenDevice calls:

```py
class SystemManager:
    """SystemManager is used by scripts to query device state and access devices."""

    async def getComponent(self, id: str) -> Any:
        pass

    def getDeviceById(self, id: str) -> ScryptedDevice:
        pass

    def getDeviceByName(self, name: str) -> ScryptedDevice:
        pass

    def getDeviceState(self, id: str) -> Any:
        pass

    def getSystemState(self) -> Any:
        pass

    def listen(self, callback: EventListener) -> EventListenerRegister:
        pass

    def listenDevice(self, id: str, event: str | EventListenerOptions, callback: EventListener) -> EventListenerRegister:
        pass

    async def removeDevice(self, id: str) -> None:
        pass
```

koush:
Tbh the eventing being a faux synchronous call is not great, I think I'd also prefer this was an async iterator pattern too. Garbage collection, throwing/returning inside an event iterator, and plugin death would trigger the unsubscription automatically.

raman325:
agreed this should switch to async. It's probably ok as is in isolation but HA wants to manage the event loop and I think faux synchronous will end up triggering some errors although I haven't tested that yet

koush:
need to undo all the typing related changes as it will not work on older versions of python. I had to revert some off our prior commits to fix crashes.

raman325:
yeah I'm moving this PR to draft, it's been a while since I've been able to work on it and I may have to start over, but I don't want to lose sight of it

raman325:
> there's also a new circular reference between types and other which may or may not cause issues, unsure.

this won't matter, it only gets triggered when type checks run which don't suffer from the circular reference problem. Your point about older versions of python is fair though

djburnside7769 (approved):

PR #1020, opened 2023-08-16
Added reolink autotrack stream + tidied loops

bendews:
Added the `autotrack` stream for Reolink Trackmix cameras, and tidied up the loops that generate the various streams to avoid accessing array indexes.

Would be nice if there was some way to detect the model and apply the ideal streams & config based on that.


koush:
Sorry, I missed this. Can't use this change because it will add an auto track stream to all Reolink Cameras, which will cause issues during stream selection, because it doesn't exist in all Reolink Cameras. 
You can create a new pull request that has stream capability detection and restricts addition of the auto track stream. Or make this an off by default setting.

PR #1034, opened 2023-08-23
core: docker installs delay update prompt until image is ready

bjia56:


plugins/core/ui/src/components/plugin/plugin.ts
```diff
@@ -54,10 +61,36 @@ export async function checkUpdate(npmPackage: string, npmPackageVersion: string)
     }
     return {
         updateAvailable,
+        updatePublished,
```
koush, line 64: it doesn't look like this field is used at all, can remove it.

PR #1044, opened 2023-08-25
0.0.11 - kick off audio and video streams sooner without waiting for …

slyoldfox:
…the SIP call to be established


koush:
On a side node, can you remove the pam-diff plugin dependency from package.json

koush:
never mind, I see that's gone

PR #1046, opened 2023-08-26
core: replace google maps with leaflet/OSM

bjia56:
Google Maps requires an API key and allow-listing of referrer urls, which causes the PositionSensor map display to not work for most domains/IPs. This change replaces it with Leaflet, using OpenStreetMap underneath.

image

Also adds a couple of babel dev dependencies since they were removed from the core babel plugin, which prevents a clean rebuild of the UI from working.


bjia56:
There's a huge diff in package-lock.json, probably because I deleted and recreated it. Let me see if I can reduce the diff by restoring the old one.

bjia56:
As discussed in Discord, the package-lock.json is ok to delete and recreate.

PR #1049, opened 2023-08-27
core: use webpack bundled map marker

bjia56:
The default Leafletjs marker icon has preconfigured values for where to place the image in relation to the pixel coordinate that matches its anchor. This default was previously ignored since the icons were missing, but the alternative did not have the correct anchor coordinates. This change restores the default marker behavior.

Also disables touch-zoom to behave more like a static image.


bjia56:
Let me know if you'd like me to bump and publish this as part of the PR.

PR #1130, opened 2023-10-13
server: add portable python

bjia56:
This is the outcome of my efforts to make Scrypted server more self-contained, since one of the largest dependencies is the Python runtime. The hope is that shipping a portable, self-contained Python installation will reduce a floating-target dependency on the host machine.

I've been using this portable Python 3.9 build on my own Scrypted installation to run Arlo, Python Codecs, and OpenVINO plugins with no noticeable issues, however it has only been validated on Linux x86_64. The proposed plan of introducing this is to make it opt-in by plugins - I would like to get the Arlo plugin moved onto the portable build to sus out any additional platform-specific issues on OSes and architectures that I cannot test, without impacting the rest of the server and plugins.


server/src/plugin/runtime/install-python.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function getPythonExecutable() {
+    const portablePython = new PortablePython(PYTHON_VERSION, pythonInstallDir());
+    return portablePython.executablePath;
+}
+
+export async function ensurePythonInstall() {
+    const portablePython = new PortablePython(PYTHON_VERSION, pythonInstallDir());
+    await portablePython.install();
```
koush, line 18: Similar to how other native binaries are packaged (ffmpeg-static, level), the install of native binaries should be done at `npm install` time via a npm `postinstall` script.

koush: this script will need to be in the portable-python npm lib

bjia56:
Closed per discussion in Discord

PR #1148, opened 2023-10-25
server & core: send terminal size info

bjia56:
Whenever I use the shell terminal for anything nontrivial, it frequently misbehaves by wrapping output in odd ways. This PR is a coordinated change between server and core by updating the engine.io/shell endpoint's interface to send annotated packets - aside from data messages, a `dim` packet can be sent to change the pty dimensions on the server side.

We can see all the screen real estate being used here:
image


koush:
Err, just took a closer look. This may be stringifying binary data. May not behave well. I think a better approach is to check if the data is a Buffer with Buffer.isBuffer on the server side.

bjia56:
Makes sense, I'll open another PR shortly

PR #1149, opened 2023-10-25
server & core: handle binary terminal data

bjia56:
Follow up to #1148 

Binary data now treated properly as such. Also optimizes the common data transfer cases by reducing the bytes sent in the stringified message and removing stringification entirely for server-sent data.

### DIFF
--- a/plugins/core/ui/src/components/builtin/ShellComponent.vue
+++ b/plugins/core/ui/src/components/builtin/ShellComponent.vue
@@ -37,16 +37,22 @@ export default {
     const rootLocation = `${window.location.protocol}//${window.location.host}`;
     this.socket = eio(rootLocation, options);
 
+    this.socket.send(JSON.stringify({ dim: { cols: term.cols, rows: term.rows } }));
+
     this.socket.on("message", (data) => {
-      term.write(new Uint8Array(Buffer.from(data)));
+      term.write(new Uint8Array(Buffer.from(JSON.parse(data).data)));
     });
 
     term.onData((data) => {
-      this.socket.send(data);
+      this.socket.send(JSON.stringify({ data }));
+    });
+
+    term.on('resize', dim => {
+      this.socket.send(JSON.stringify({ dim }));
     });
 
     term.onBinary((data) => {
-      this.socket.send(data);
+      this.socket.send(JSON.stringify({ data }));
     });
   },
   destroyed() {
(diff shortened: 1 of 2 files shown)


plugins/core/ui/src/components/builtin/ShellComponent.vue
```diff
@@ -40,19 +40,20 @@ export default {
     this.socket.send(JSON.stringify({ dim: { cols: term.cols, rows: term.rows } }));
 
     this.socket.on("message", (data) => {
-      term.write(new Uint8Array(Buffer.from(JSON.parse(data).data)));
+      term.write(new Uint8Array(Buffer.from(data)));
     });
 
     term.onData((data) => {
-      this.socket.send(JSON.stringify({ data }));
+      this.socket.send(JSON.stringify({ d: data }));
```
koush, line 47: Should send this as a buffer as well, all term data can go across the wire as binary. No reason to send json for text.

server/src/runtime.ts
```diff
@@ -119,11 +119,16 @@ export class ScryptedRuntime extends PluginHttp<HttpPluginData> {
                 const spawn = require('node-pty-prebuilt-multiarch').spawn as typeof ptySpawn;
                 const cp = spawn(process.env.SHELL, [], {
                 });
-                cp.onData(data => connection.send(JSON.stringify({data})));
+                cp.onData(data => connection.send(data));
                 connection.on('message', message => {
+                    if (Buffer.isBuffer(message)) {
+                        cp.write(message.toString());
```
koush, line 125: Why toString here, can write the buffer?

server/src/runtime.ts
```diff
@@ -119,11 +119,16 @@ export class ScryptedRuntime extends PluginHttp<HttpPluginData> {
...
+                        return;
+                    }
+
                     const parsed = JSON.parse(message.toString());
-                    if (parsed.data) {
-                        cp.write(parsed.data);
+                    if (parsed.d) {
```
koush, line 130: This code path will be gone with aforementioned changes

server/src/runtime.ts
bjia56: According to this, they are expecting a string https://github.com/microsoft/node-pty/blob/main/typings/node-pty.d.ts#L168

koush: Huh, weird. Looks good then

bjia56:
With the latest changes, core is functionally backwards compatible with existing servers, except it will type in the dimensions into the pty. Server is not backwards compatible with existing core (the console would just stop working, but at least won't crash the server), so it's probably better to release core first, then server shortly afterwards.

server/src/runtime.ts
```diff
@@ -127,9 +127,7 @@ export class ScryptedRuntime extends PluginHttp<HttpPluginData> {
                     }
 
                     const parsed = JSON.parse(message.toString());
```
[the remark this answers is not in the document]
koush: Oh weird, I must have been looking at an old revision or something, cause it's definitely there now!

PR #1153, opened 2023-10-26
cli, client, server: add interactive `npx scrypted shell`

bjia56:
This change adds a new command, `npx scrypted shell [localhost[:10443]]` which will connect the client's TTY to the Scrypted server engine.io/shell endpoint. The server has also been updated to close the engine.io connection when the remote shell exits. This changed server behavior does not impact core plugin (web terminal client), and is used to terminate the `npx scrypted shell` process.

I ran into some dependency issues when trying to build `cli`, namely it looks like `client` is missing the `engine.io` package. Perhaps it needs to be added before publishing.

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -49,6 +49,7 @@ export interface ScryptedClientStatic extends ScryptedStatic {
     connectionType: ScryptedClientConnectionType;
     rpcPeer: RpcPeer;
     loginResult: ScryptedClientLoginResult;
+    connectShell(): IOClientSocket;
 }
 
 export interface ScryptedConnectionOptions {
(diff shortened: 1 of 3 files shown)


packages/client/src/index.ts
```diff
@@ -736,6 +740,22 @@ export async function connectScryptedClient(options: ScryptedClientOptions): Pro
                 queryToken,
                 authorization,
                 cloudAddress,
+            },
+            connectShell() {
```
koush, line 744: I don't think adding a new method specifically for a shell here is necessary. Maybe what you want is extraHeaders?

koush: Actually I think you can append the queryToken stuff that's already there to authorize any url like engine.io shell endpoint.

bjia56:
Closing this in favor of a new socket RPC and async iterator approach discussed over Discord

PR #1166, opened 2023-11-06
server, client: connectRPCObject for web api clients

bjia56:
Uses a new engine.io endpoint to serve as the transport for connectRPCObject requests coming from the web client.

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -749,7 +749,7 @@ export async function connectScryptedClient(options: ScryptedClientOptions): Pro
                         const serializer = createRpcDuplexSerializer({
                             write: data => clusterPeerSocket.send(data),
                         });
-                        clusterPeerSocket.on('message', data => serializer.onData(Buffer.from(data)));
+                        clusterPeerSocket.on('message', data => serializer.onData(data));
 
                         const clusterPeer = new RpcPeer(clientName || 'engine.io-client', "cluster-proxy", (message, reject, serializationContext) => {
                             try {
@@ -759,7 +759,7 @@ export async function connectScryptedClient(options: ScryptedClientOptions): Pro
                                 reject?.(e);
                             }
                         });
-                        serializer.setupRpcPeer(rpcPeer);
+                        serializer.setupRpcPeer(clusterPeer);
                         clusterPeer.tags.localPort = 65535 + 1; // beyond max port number
                         peerReady = true;
                         return { clusterPeer, clusterSecret };
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -12,7 +12,7 @@ import { MediaObject } from '../../../server/src/plugin/mediaobject';
 import { attachPluginRemote } from '../../../server/src/plugin/plugin-remote';
 import { ClusterObject, ConnectRPCObject } from '../../../server/src/plugin/plugin-remote-worker';
 import { RpcPeer } from '../../../server/src/rpc';
-import { createRpcDuplexSerializer, createRpcSerializer } from '../../../server/src/rpc-serializer';
+import { createDuplexRpcPeer, createRpcDuplexSerializer, createRpcSerializer } from '../../../server/src/rpc-serializer';
 import packageJson from '../package.json';
 import { isIPAddress } from "./ip";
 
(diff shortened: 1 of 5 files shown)


packages/client/src/index.ts
```diff
@@ -9,11 +10,14 @@ import { DataChannelDebouncer } from "../../../plugins/webrtc/src/datachannel-de
 import type { IOSocket } from '../../../server/src/io';
 import { MediaObject } from '../../../server/src/plugin/mediaobject';
 import { attachPluginRemote } from '../../../server/src/plugin/plugin-remote';
+import { ClusterObject, ConnectRPCObject } from '../../../server/src/plugin/plugin-remote-worker';
 import { RpcPeer } from '../../../server/src/rpc';
 import { createRpcDuplexSerializer, createRpcSerializer } from '../../../server/src/rpc-serializer';
 import packageJson from '../package.json';
 import { isIPAddress } from "./ip";
 
+const connectRPCObjectPlaceholderPort = 65535 + 1; // beyond max port number
```
koush, line 19: This should be unique to this client so as to not cause conflicts between multiple web clients.

koush: Does not need to be a number. I think the signature should be come sourcePeerId

packages/client/src/index.ts
```diff
@@ -707,6 +711,96 @@ export async function connectScryptedClient(options: ScryptedClientOptions): Pro
...
+                    const eioEndpoint = baseUrl ? new URL(eioPath, baseUrl).pathname : '/' + eioPath;
+                    const clusterPeerOptions = {
+                        path: eioEndpoint,
+                        query: {
+                            cacehBust,
+                        },
+                        withCredentials: true,
+                        extraHeaders: {
+                            ...extraHeaders,
+                            "x-scrypted-clusterobject-port": `${port}`,
```
koush, line 729: send this as a query parameter. server parse/require it.

packages/client/src/index.ts
```diff
@@ -707,6 +711,96 @@ export async function connectScryptedClient(options: ScryptedClientOptions): Pro
...
+
+        const connectRPCObject = async (value: any) => {
+            const clusterObject: ClusterObject = value?.__cluster;
+            const { port, proxyId, source } = clusterObject;
+
+            try {
+                const clusterPeerPromise = ensureClusterPeer(port);
+                const { clusterPeer, clusterSecret } = await clusterPeerPromise;
+                // this object is already connected
+                if (clusterPeer.tags.localPort === connectRPCObjectPlaceholderPort)
```
koush, line 789: The check should be `source` not `connectRPCObjectPlaceholderPort`.

koush: Actually this check can be removed altogether. This client will never create objects that are reachable on demand, because there's no way to reach this client on demand.

packages/client/src/index.ts
```diff
@@ -707,6 +711,96 @@ export async function connectScryptedClient(options: ScryptedClientOptions): Pro
...
+                })();
+                clusterPeers.set(port, clusterPeerPromise);
+            }
+            return clusterPeerPromise;
+        };
+
+
+        const connectRPCObject = async (value: any) => {
+            const clusterObject: ClusterObject = value?.__cluster;
+            const { port, proxyId, source } = clusterObject;
```
koush, line 783: clusterObject may be undefined.

koush: the resolveObject function is missing here too, that is used to prevent connect from being called multiple times on the same object resulting in duplication.

packages/client/src/index.ts
koush: actually:

```ts
                if (clusterPeer.tags.localPort === source)
```

is the correct check (as per upstream)

koush:
uhh disregard that change, didn't realize I was working in your branch.

koush:
ok backed it out with a force push, pretty sure this was the state before I clobbered your stuff.

packages/client/src/index.ts
```diff
@@ -707,6 +711,110 @@ export async function connectScryptedClient(options: ScryptedClientOptions): Pro
             .map(id => systemManager.getDeviceById(id))
             .find(device => device.pluginId === '@scrypted/core' && device.nativeId === `user:${username}`);
 
+        const clusterPeers = new Map<number, Promise<{ clusterPeer: RpcPeer, clusterSecret: string }>>();
+        const ensureClusterPeer = (port: number) => {
+            let clusterPeerPromise = clusterPeers.get(port);
+            if (!clusterPeerPromise) {
+                clusterPeerPromise = (async () => {
+                    const eioPath = 'engine.io/connectRPCObject';
+                    const eioEndpoint = baseUrl ? new URL(eioPath, baseUrl).pathname : '/' + eioPath;
```
koush, line 720: Should use relative paths. Home assistant will rehost Scrypted's web interface on a sub path.

packages/client/src/index.ts
```diff
@@ -9,11 +10,14 @@ import { DataChannelDebouncer } from "../../../plugins/webrtc/src/datachannel-de
 import type { IOSocket } from '../../../server/src/io';
 import { MediaObject } from '../../../server/src/plugin/mediaobject';
 import { attachPluginRemote } from '../../../server/src/plugin/plugin-remote';
+import { ClusterObject, ConnectRPCObject } from '../../../server/src/plugin/plugin-remote-worker';
```
koush, line 13: This import pulls in a whole bunch of dependencies. You must use `import type` to prevent that. Ideally, these types maybe ought to live in their own .ts file.

server/src/runtime.ts
```diff
@@ -114,7 +126,7 @@ export class ScryptedRuntime extends PluginHttp<HttpPluginData> {
             this.shellHandler(req, res);
         });
 
-        this.shellio.on('connection', connection => {
+        this.shellIO.on('connection', connection => {
```
koush, line 129: For the sake of a minimal diff, leave this casing be. We'll be removing this endpoint anyways.

packages/client/src/index.ts
```diff
@@ -9,11 +10,14 @@ import { DataChannelDebouncer } from "../../../plugins/webrtc/src/datachannel-de
...
 import { RpcPeer } from '../../../server/src/rpc';
 import { createRpcDuplexSerializer, createRpcSerializer } from '../../../server/src/rpc-serializer';
 import packageJson from '../package.json';
 import { isIPAddress } from "./ip";
 
+const sourcePeerId = [...new Array(8)].map(() => RpcPeer.RANDOM_DIGITS.charAt(Math.floor(Math.random() * RpcPeer.RANDOM_DIGITS.length))).join('');
```
koush, line 19: Could make RpcPeer.generateId() static, and then call that rather than copy this bit.

server/src/plugin/connect-rpc-object.ts
```diff
@@ -0,0 +1,30 @@
...
+    source: number;
+}
+
+export type ConnectRPCObject = (id: string, secret: string, sourcePeerPort: number) => Promise<any>;
+
+/*
+ * Handle incoming connections that will be
+ * proxied to a connectRPCObject socket.
+ */
+export function setupConnectRPCObjectProxy(clusterSecret: string, port: number, connection: Socket & IOSocket) {
```
koush, line 18: I think you mean Socket | IOSocket

PR #1170, opened 2023-11-09
server, client: send full ClusterObject on new eio endpoint

bjia56:
Also changes the checksums computed on the python side to be symmetric to node's signatures (both now use base64 encoding).


server/src/runtime.ts
```diff
@@ -163,7 +163,18 @@ export class ScryptedRuntime extends PluginHttp<HttpPluginData> {
                 res.end();
                 return;
             }
-            if (!req.query.port) {
+            if (!req.query.clusterObject) {
```
koush, line 166: couple things I'm just now noticing:

the act check above is not necessary, non-admin users should be able to access this API.

the cluster object validity check should be done below when the connection is initiated. return some error json and terminate the connection there.

I think this whole handler can then be removed.

bjia56: Ok, sounds good will update

PR #1171, opened 2023-11-11
client, server, core: Add StreamService interface and TerminalService device

bjia56:
- Add a new StreamService interface to connect bidirectional streams with AsyncGenerators
- Add a new TerminalService device to core plugin to replace existing server shell functionality
- Update client, ui to use new TerminalService device over connectRPCObject


common/src/async-queue.ts
```diff
@@ -34,6 +35,15 @@ export function createAsyncQueue<T>() {
             throw ended;
     }
 
+    const peek = () => {
+        if (queued.length) {
```
koush, line 39: No changes should be necessary in this queue, calling clear or dequeue is sufficient. Catch any errors if you need to..

plugins/core/src/terminal-service.ts
```diff
@@ -0,0 +1,43 @@
...
+import { BufferedBuffer } from '@scrypted/common/src/buffered-buffer';
+
+export const TerminalServiceNativeId = 'terminalservice';
+
+export class TerminalService extends ScryptedDeviceBase implements StreamService {
+    async connectStream(input: AsyncGenerator<any, void>): Promise<AsyncGenerator<any, void>> {
+        const spawn = require('node-pty-prebuilt-multiarch').spawn as typeof ptySpawn;
+        const cp: IPty = spawn(process.env.SHELL, [], {});
+        const buffer = new BufferedBuffer();
+        buffer.onClose(() => cp.kill());
```
koush, line 12: This is unnecessary and also racey. on end will be called before the queue is flushed.

common/src/buffered-buffer.ts
```diff
@@ -0,0 +1,55 @@
...
+    onClose(fn: Function) {
+        this.queue.onEnd(fn);
+    }
+
+    /**
+     * The returned AsyncGenerator will concatenate together
+     * sequential Buffers, but leave strings unchanged.
+     */
+    async *generator(): AsyncGenerator<Buffer | string, void> {
+        while (!this.queue.ended) {
```
koush, line 29: I think what you want here is clear, followed by a concat.

koush: If clear results in empty set, then await dequeue

common/src/buffered-buffer.ts
```diff
@@ -0,0 +1,55 @@
+import { createAsyncQueue } from "./async-queue"
+
+const MAX_BUFFERED_BUFFER = 100;
```
koush, line 3: It’s possible to use a running total. When you awaiting an enqueue lets you know when it is dequeued. Can maintain a tally.

100 1 byte packets would trigger a disconnect here.

common/src/buffered-buffer.ts
bjia56: Reasoning for doing it this way iteratively is to keep string messages intact but concat all buffers. A clear would then need an iteration through the list to find if there's a string, concat elements before the string, then requeue the rest of the list

koush: only the client side sends strings I think. it would probably be easier to follow by removing this this abstracted bufferedbuffer altogether and use the async queue directly on both sides.

plugins/core/src/terminal-service.ts
```diff
@@ -0,0 +1,73 @@
...
+
+                    yield await queue.dequeue();
+                }
+            }
+            finally {
+                cp.kill();
+            }
+        }
+
+        setTimeout(async () => {
```
koush, line 45: I think I had this setTimeout pattern before due to some weird issue with the event loop on engine.io. I think this can be replaced with an in place

(async() => {
 ... code
})()

plugins/core/src/terminal-service.ts
```diff
@@ -0,0 +1,73 @@
...
+                        const parsed = JSON.parse(message.toString());
+                        if (parsed.dim) {
+                            cp.resize(parsed.dim.cols, parsed.dim.rows);
+                        }
+                    } catch {
+                        cp.write(message.toString());
+                    }
+                }
+            }
+            finally {
```
koush, line 66: Should catch and eat all errors here.

plugins/core/src/terminal-service.ts
```diff
@@ -0,0 +1,73 @@
...
+import { createAsyncQueue } from '@scrypted/common/src/async-queue'
+
+export const TerminalServiceNativeId = 'terminalservice';
+
+export class TerminalService extends ScryptedDeviceBase implements StreamService {
+    async connectStream(input: AsyncGenerator<any, void>): Promise<AsyncGenerator<any, void>> {
+        const spawn = require('node-pty-prebuilt-multiarch').spawn as typeof ptySpawn;
+        const cp: IPty = spawn(process.env.SHELL, [], {});
+        const queue = createAsyncQueue<Buffer>();
+        cp.onExit(() => queue.end());
```
koush, line 12: does cp have an error event? may need to listen and call queue.end(error) in that case?

koush:
looking good! had a few remaining questions, and remembered that all errors need be caught in the unmonitored promise inside the setTiemout.

plugins/core/src/terminal-service.ts
bjia56: It doesn't look like it, the typing here: https://github.com/microsoft/node-pty/blob/main/typings/node-pty.d.ts#L108

PR #1174, opened 2023-11-13
cli, core, ui: add `npx scrypted shell` and support interactive/noninteractive shells

bjia56:
This is a reimplementation of #1153, but following the new design paradigm of using connectRPCObject and buffering data.

~~Unfortunately, I'm not able to get non-interactive shell working for the current setup - testing `echo date | node ./dist/packages/cli/src/main.js shell localhost:10443` closes the connection immediately without any output. Probably needs a change to core plugin to keep the connection half-open if one generator ends before the other, but that's a bit tricky to do since it's unclear when to terminate the child process. In my opinion, that part can wait for later as an enhancement, since interactive shell is going to be the main use case.~~
^ noninteractive use has been implemented in 30db66d2d353dce6b0817bd94c1c5575b5cd090a

### DIFF
--- a/packages/cli/src/main.ts
+++ b/packages/cli/src/main.ts
@@ -263,7 +263,6 @@ async function main() {
                     process.stdin.resume();
             }
         });
-        process.stdin.end(() => queue.end());
 
         async function* generator() {
             try {
@@ -276,12 +275,19 @@ async function main() {
                         // groups of buffer data, but allow string control messages
                         // to be sent without any batching
                         for (let i = 0; i < buffers.length; ++i) {
-                            if (!Buffer.isBuffer(buffers[i]) && i != buffersStart) {
-                                yield Buffer.concat(buffers.slice(buffersStart, i) as Buffer[]);
+                            if (!Buffer.isBuffer(buffers[i])) {
+                                if (i != buffersStart) {
+                                    yield Buffer.concat(buffers.slice(buffersStart, i) as Buffer[]);
+                                }
+                                buffersStart = i + 1;
+                                yield buffers[i];
                             }
-                            buffersStart = i + 1;
-                            yield buffers[i];
                         }
+
+                        if (buffersStart != buffers.length) {
+                            yield Buffer.concat(buffers.slice(buffersStart, buffers.length) as Buffer[]);
+                        }
+
                         continue;
                     }
 
--- a/packages/cli/src/main.ts
+++ b/packages/cli/src/main.ts
@@ -11,7 +11,7 @@ import { connectScryptedClient } from '@scrypted/client';
 import { ScryptedMimeTypes, FFmpegInput, DeviceProvider, StreamService } from '@scrypted/types';
 import semver from 'semver';
 import child_process from 'child_process';
-import { createAsyncQueue } from '../../../common/src/async-queue';
+import { connectShell } from './shell';
 
 const httpsAgent = new https.Agent({
     rejectUnauthorized: false,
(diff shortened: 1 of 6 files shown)


packages/cli/src/main.ts
```diff
@@ -220,6 +221,104 @@ async function main() {
 
         console.log('install successful. id:', response.data.id);
     }
+    else if (process.argv[2] === 'shell') {
```
koush, line 224: Should split this out into a separate file

packages/cli/src/main.ts
```diff
@@ -220,6 +221,104 @@ async function main() {
...
+        });
+
+        async function* generator() {
+            try {
+                while (true) {
+                    const buffers = queue.clear();
+                    if (buffers.length) {
+                        let buffersStart = 0;
+
+                        // this wonky loop-concat-yield is to batch together
```
koush, line 276: This could be simplified significantly with 2 queues. Control messages put the message in the control queue and an empty buffer in the buffer queue. Clear control message queue first, send that if a message exists.

koush:
Something seems off if the cp is immediately exiting. Queue end should be called after all data has been flushed. Seems it’s called prematurely. Let me check pty events.

packages/cli/src/main.ts
```diff
@@ -220,6 +221,104 @@ async function main() {
...
+            for await (const message of await termSvcDirect.connectStream(generator())) {
+                if (!message) {
+                    process.exit();
+                }
+                process.stdout.write(new Uint8Array(Buffer.from(message)));
+            }
+        } catch {
+            // ignore
+        } finally {
+            process.exit();
```
koush, line 319: I think this exit is causing the premature exit. Should probably exit only when the remote generator ends.

bjia56: Removing this didn't quite help. I need to send additional data to the remote since the remote shell should behave differently for noninteractive use.

packages/cli/src/shell.ts
```diff
@@ -0,0 +1,90 @@
...
+                            yield concat;
+                        }
+                    }
+                    yield JSON.stringify(ctrl);
+                }
+                continue;
+            }
+
+            const dataBuffers = dataQueue.clear();
+            const concat = Buffer.concat(dataBuffers);
```
koush, line 59: feels like a race or spin condition here as this doesn't mirror the other code. should be something like:

```ts
const dataBuffers = dataQueue.clear();
if (dataBuffers.length === 0) {
  const buf = await dataQueue.dequeue();
   if (buf.length)
       yield buf;
  continue;
}

const concat = Buffer.concat(dataBuffers);
if (concat.length)
  yield concat;
```

packages/cli/src/shell.ts
```diff
@@ -0,0 +1,90 @@
...
+    });
+
+    async function* generator() {
+        while (true) {
+            const ctrlBuffers = ctrlQueue.clear();
+            if (ctrlBuffers.length) {
+                for (const ctrl of ctrlBuffers) {
+                    if (ctrl.eof) {
+                        // flush the buffer before sending eof
+                        const dataBuffers = dataQueue.clear();
```
koush, line 47: Oh, I hadn't considered the message ordering mattering between data and control. was this added for non-pty pipe handling?

bjia56: Yeah since the ctrl messages take priority, we should check that any data in the buffer should be flushed for EOF case, since otherwise it will not get processed in the remote after the EOF message

PR #1241, opened 2024-01-03
Switch Reolink Doorbell from API person detection to Onvif Person detection

minijaws:
The person detection event from the doorbell is also available from Onvif. I have switched to using that instead so it is not required to poll the API every 1 second for detection


koush:
Hi, thanks for this change. Unfortunately not all reolink cameras are guaranteed to have onvif. I'm not sure this is a safe change to make.

koush:
done, though I used a different approach to not modify onvif on all cameras.

PR #1273, opened 2024-01-18
Implemented other intercom codec for Dahua doorbells (G.711A)

JBosecker:
The intercom was not working with my VTO2000, so I dug into the problem and found out, that only G.711A is supported.
So if `Two Way Audio` is `Amcrest` and `Doorbell Type` is `Dahua Doorbell` G.711A is used instead of AAC.

### DIFF
--- a/plugins/amcrest/src/main.ts
+++ b/plugins/amcrest/src/main.ts
@@ -495,6 +495,8 @@ class AmcrestCamera extends RtspSmartCamera implements VideoCameraConfiguration,
             return this.onvifIntercom.startIntercom(media);
         }
 
+        const doorbellType = this.storage.getItem('doorbellType');
+        
         // not sure if this all works, since i don't actually have a doorbell.
         // good luck!
         const channel = this.getRtspChannel() || '1';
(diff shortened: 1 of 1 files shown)


koush:
awesome. is there a way to determine the codec via their api? how did you figure it out? I don't see a way at first glance:

https://cdn.discordapp.com/attachments/889911243853418547/895148494459908157/AmcrestHTTPAPI9.2020.pdf?ex=65b9bc80&is=65a74780&hm=c2f1cca00583703a4ac89adf7cb63f05c8b5966c1f100f7f5750c96079e31739&

JBosecker:
Sadly I didn't find a way...
They were using G.711A in their examples, so I tried that.
One could think that other codecs are supported while reading the documentation, but at least for VTO2000 it's not the case.
I cannot be sure, but I think G.711A should be more compatible than AAC.

koush:
incidentally does the device support onvif two way?

JBosecker:
Thanks for merging the PR! :-)

No, ONVIF two way audio does not seem to work:
`[HomeKit]: intercom failed to start RPCResultError: auth failed
[HomeKit]:     at RtspClient.request (/common/src/rtsp-server.ts:613:23)
[HomeKit]:     at processTicksAndRejections (node:internal/process/task_queues:95:5)
[HomeKit]:     at RtspClient.options (/common/src/rtsp-server.ts:628:21)
[HomeKit]:     at OnvifIntercom.checkIntercom (/onvif/src/onvif-intercom.ts:30:9)
[HomeKit]:     at OnvifIntercom.startIntercom (/onvif/src/onvif-intercom.ts:50:54)
[HomeKit]:     at RpcPeer.handleMessageInternal (/server/node_modules/@scrypted/server/src/rpc.ts:759:37)
[HomeKit]: @scrypted/amcrest:host
[HomeKit]: host:@scrypted/homekit {
[HomeKit]:   cause: undefined
[HomeKit]: }`

koush:
you may need to enable onvif support and create an onvif user

JBosecker:
I just walked through the web interface and there is no option for ONVIF :-/

PR #1275, opened 2024-01-21
server: expose backup as a service

bjia56:
Moves the backup endpoint logic to a service that can be accessed by plugins with `systemManager.getComponent('backup')`.


koush:
let's add restore too

PR #1295, opened 2024-02-04
Periodic snapshot timeout

longzheng:
I noticed after the snapshot timeout was added in https://github.com/koush/scrypted/commit/0e700a53d0a673e336cdaca7e0265ed3bd30ae2e, my Homekit snapshots were always delayed (the update would show the previous update maybe 70% of the time).

After doing some debugging, I noticed that my specific set up (Dahua/Amcrest 6MP cameras) would always take around a fraction of a second longer (e.g. 1.4 seconds) to return the snapshots. (I've always tried adjusting the snapshot quality but it didn't make a big enough difference). I assume this is just the nature of 6MP snapshots.

I believe the 1 second timeout was added to minimize any delays it may cause HomeKit which is understandable, however I'm willing to sacrifice a marginal delay for more up-to-date snapshots, especially over mobile where the periodic snapshots only update once every 30 seconds.

I thought it would be best if this was a configurable setting with a default of 1 second. 

For me 2 seconds was enough to avoid the issue, and it doesn't appear to have a significant impact on the HomeKit experience.

image

I also added some debug logging of how long the periodic snapshot takes and if it takes longer than the configured timeout, so people can easily debug why their snapshots might be old.

```
[Backyard] Periodic snapshot took 1.421 seconds to retrieve.
[Backyard] Periodic snapshot took longer than 1 seconds to retrieve, using previous snapshot.
```

### DIFF
--- a/plugins/snapshot/src/main.ts
+++ b/plugins/snapshot/src/main.ts
@@ -293,11 +293,11 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
                 const cp = this.currentPicture;
                 debounced.catch(() => {});
                 try {
-                    picture = await (options.periodicTimeout ? timeoutPromise(options.periodicTimeout, debounced) : debounced);
+                    picture = await (options.timeout ? timeoutPromise(options.timeout, debounced) : debounced);
                 }
                 catch (e) {
-                    if (options.periodicTimeout)
-                        this.debugConsole.debug(`Periodic snapshot took longer than ${options.periodicTimeout} seconds to retrieve, falling back to cached picture.`)
+                    if (options.timeout)
+                        this.debugConsole.debug(`Periodic snapshot took longer than ${options.timeout} seconds to retrieve, falling back to cached picture.`)
 
                     picture = cp;
                 }
--- a/plugins/snapshot/src/main.ts
+++ b/plugins/snapshot/src/main.ts
@@ -296,6 +296,9 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
                     picture = await (options.periodicTimeout ? timeoutPromise(options.periodicTimeout, debounced) : debounced);
                 }
                 catch (e) {
+                    if (options.periodicTimeout)
+                        this.debugConsole.debug(`Periodic snapshot took longer than ${options.periodicTimeout} seconds to retrieve, falling back to cached picture.`)
+
                     picture = cp;
                 }
             }
--- a/plugins/snapshot/src/main.ts
+++ b/plugins/snapshot/src/main.ts
@@ -293,7 +293,7 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
                 const cp = this.currentPicture;
                 debounced.catch(() => {});
                 try {
-                    picture = await timeoutPromise(1000, debounced);
+                    picture = await (options.periodicTimeout ? timeoutPromise(options.periodicTimeout, debounced) : debounced);
                 }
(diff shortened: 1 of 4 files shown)


longzheng:
After discussion on Discord [#1](https://discord.com/channels/882329362295316500/886111134367772673/1203739758715600936) [#2](https://discord.com/channels/882329362295316500/886111134367772673/1203776871444127744), I changed the approach so that `RequestPictureOptions` has a `periodicTimeout` which HomeKit sets to 2 seconds.

For other create snapshot usages there is no timeout.

sdk/types/src/types.input.ts
```diff
@@ -453,6 +453,10 @@ export interface RequestPictureOptions extends PictureOptions {
    * Flag that hints whether multiple cameras are being refreshed by this user request. Can be used to prefetch the snapshots.
    */
   bulkRequest?: boolean;
+  /**
+   * The maximum time in milliseconds to wait for the periodic picture to be taken.
+   */
+  periodicTimeout?: number;
```
koush, line 459: can you rename this to timeout, I'll use the property on all requests regardless of snapshot reason

longzheng: Fair enough, updated.

plugins/snapshot/src/main.ts
```diff
@@ -291,9 +293,12 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
                 const cp = this.currentPicture;
                 debounced.catch(() => {});
                 try {
-                    picture = await timeoutPromise(1000, debounced);
+                    picture = await (options.timeout ? timeoutPromise(options.timeout, debounced) : debounced);
                 }
                 catch (e) {
+                    if (options.timeout)
+                        this.debugConsole.debug(`Periodic snapshot took longer than ${options.timeout} seconds to retrieve, falling back to cached picture.`)
```
longzheng, line 300: @koush I was thinking this should maybe a `.log()` instead of `.debug()` since it might be helpful to people who don't have debug enabled to easily see why their snapshots are stale.

koush: I think I meant for it to be log actually, so sure

longzheng:
@koush just double checking if you had any more thoughts on this.

I experience this HomeKit quirk every day so would be keen to see it merged and published 😇

plugins/snapshot/src/main.ts
```diff
@@ -291,9 +293,12 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
                 const cp = this.currentPicture;
                 debounced.catch(() => {});
                 try {
-                    picture = await timeoutPromise(1000, debounced);
+                    picture = await (options.timeout ? timeoutPromise(options.timeout, debounced) : debounced);
```
koush, line 296: regression here, as there is no timeout value provided. periodic snapshots must have a default timeout if none is provided. since it is periodic, erroring is fine, as the upstream will be updating the snapshot regularly and recover from errors. cache usage is also acceptable.

koush: @longzheng

longzheng: Ah, I wonder if the approach should be to have a default here

```ts
                    picture = await timeoutPromise(options.timeout ? options.timeout: 1000, debounced);
```

Or have `options` default `timeout` to `1000` elsewhere?

koush: Yep that's what I did. I dont think this should affect the initial issue of homekit having stale snapshots, but figured I'd bring it up. Change is already published.

PR #1301, opened 2024-02-06
webhook: minor consistency fixes

bjia56:


plugins/webhook/src/main.ts
```diff
@@ -212,6 +219,10 @@ class WebhookPlugin extends ScryptedDeviceBase implements Settings, MixinProvide
     }
 
     async getMixin(mixinDevice: any, mixinDeviceInterfaces: ScryptedInterface[], mixinDeviceState: { [key: string]: any; }): Promise<any> {
+        const id = mixinDeviceState.id;
+        if (this.createdMixins.has(id)) {
+            return this.createdMixins.get(id);
```
koush, line 224: this is incorrect, the underlying mixinDevice may have changed due to plugin reload or device reload. should return a new instance.

plugins/webhook/src/main.ts
```diff
@@ -220,13 +231,14 @@ class WebhookPlugin extends ScryptedDeviceBase implements Settings, MixinProvide
...
         });
-
-        this.createdMixins.set(ret.id, ret);
+        this.createdMixins.set(id, ret);
         return ret;
     }
 
     async releaseMixin(id: string, mixinDevice: any): Promise<void> {
-        this.createdMixins.delete(id);
+        if (this.createdMixins.get(id) === mixinDevice) {
```
koush, line 239: my hunch is this is the actual correct fix

koush:
when I looked at this code last, I recall thinking that the mixin instance wasn't even necessary, as the device state was never changed. simply using systemManager.getDeviceById would avoid this entire class of issues I believe.

bjia56:
Makes sense, though I'm not too confident in my understanding of the mixin class to do that kind of refactoring in this PR

kiler129:
According to my logs no 404s since yesterday, so cautiously I think it may be fixed 🎉  It doesn't seem like it broke anything either so it seems safe ;)

PR #1308, opened 2024-02-11
Add strict types to sdk

longzheng:
Enable TypeScript strict for `sdk/types`

Enable TypeScript strict strict for `sdk`

Fix various type-related issues

Add `build-sdk` GitHub Actions workflow to ensure SDK builds and doesn't have errors

### DIFF
--- a/sdk/src/index.ts
+++ b/sdk/src/index.ts
@@ -9,29 +9,10 @@ export class ScryptedDeviceBase extends DeviceBase {
   private _storage: Storage | undefined;
   private _log: Logger | undefined;
   private _console: Console | undefined;
-  private _deviceState: DeviceState | undefined;
+  public _deviceState: DeviceState | undefined;
 
   constructor(public readonly nativeId?: string) {
     super();
-
-    addScryptedInterfaceProperties({
-      thisObject: this,
-      getSet: (field: keyof DeviceState) => {
-        return (value: any) => {
-          this._lazyLoadDeviceState();
-          if (!this._deviceState)
-            console.warn('device state is unavailable. the device must be discovered with deviceManager.onDeviceDiscovered or deviceManager.onDevicesChanged before the state can be set.');
-          else
-            this._deviceState[field] = value;
-        }
-      },
-      getGet: (field: keyof DeviceState) => {
-        return () => {
-          this._lazyLoadDeviceState();
-          return this._deviceState?.[field];
-        }
-      }
-    })
   }
 
   get storage() {
@@ -109,7 +90,7 @@ export interface MixinDeviceOptions<T> {
   private mixinStorageSuffix: string | undefined;
   private _log: Logger | undefined;
   private _console: Console | undefined;
-  private _deviceState: WritableDeviceState;
+  public _deviceState: WritableDeviceState;
   private _listeners = new Set<EventListenerRegister>();
 
   constructor(options: MixinDeviceOptions<T>) {
@@ -128,25 +109,6 @@ export interface MixinDeviceOptions<T> {
       this._deviceState = deviceManager.createDeviceState(this._deviceState.id, this._deviceState.setState);
     }
     this.mixinProviderNativeId = options.mixinProviderNativeId;
-
-    addScryptedInterfaceProperties({
-      thisObject: this,
-      getSet: (field: keyof DeviceState) => {
-        return (value: any) => {
-          this._lazyLoadDeviceState();
-          if (!this._deviceState)
-            console.warn('device state is unavailable. the device must be discovered with deviceManager.onDeviceDiscovered or deviceManager.onDevicesChanged before the state can be set.');
-          else
-            this._deviceState[field] = value;
-        }
-      },
-      getGet: (field: keyof DeviceState) => {
-        return () => {
-          this._lazyLoadDeviceState();
-          return this._deviceState[field];
-        }
-      }
-    })
   }
 
   get storage() {
@@ -222,6 +184,38 @@ function addScryptedInterfaceProperties<T>(
     });
   }
 }
+
+(function () {
+  function _createGetState<T>(deviceBase: ScryptedDeviceBase | MixinDeviceBase<T>, state: keyof DeviceState) {
+    return function () {
+      deviceBase._lazyLoadDeviceState();
+      return deviceBase._deviceState?.[state];
+    };
+  }
+
+  function _createSetState<T>(deviceBase: ScryptedDeviceBase | MixinDeviceBase<T>, state: keyof DeviceState) {
+    return function (value: any) {
+      deviceBase._lazyLoadDeviceState();
+      if (!deviceBase._deviceState)
+        console.warn('device state is unavailable. the device must be discovered with deviceManager.onDeviceDiscovered or deviceManager.onDevicesChanged before the state can be set.');
+      else
+        deviceBase._deviceState[state] = value;
+    };
+  }
+
+  for (const field of Object.values(ScryptedInterfaceProperty)) {
+    if (field === ScryptedInterfaceProperty.nativeId)
+      continue;
+    Object.defineProperty(ScryptedDeviceBase.prototype, field, {
+      set: _createSetState(ScryptedDeviceBase.prototype, field),
+      get: _createGetState(ScryptedDeviceBase.prototype, field),
+    });
+    Object.defineProperty(MixinDeviceBase.prototype, field, {
+      set: _createSetState(MixinDeviceBase.prototype, field),
+      get: _createGetState(MixinDeviceBase.prototype, field),
+    });
+  }
+})();
 
 export const sdk: ScryptedStatic = {} as any;
 declare const deviceManager: DeviceManager;
--- a/sdk/src/index.ts
+++ b/sdk/src/index.ts
@@ -13,6 +13,25 @@ export class ScryptedDeviceBase extends DeviceBase {
 
   constructor(public readonly nativeId?: string) {
     super();
+
+    addScryptedInterfaceProperties({
+      thisObject: this,
+      getSet: (field: keyof DeviceState) => {
+        return (value: any) => {
+          this._lazyLoadDeviceState();
+          if (!this._deviceState)
+            console.warn('device state is unavailable. the device must be discovered with deviceManager.onDeviceDiscovered or deviceManager.onDevicesChanged before the state can be set.');
+          else
+            this._deviceState[field] = value;
+        }
+      },
+      getGet: (field: keyof DeviceState) => {
+        return () => {
+          this._lazyLoadDeviceState();
+          return this._deviceState?.[field];
+        }
+      }
+    })
   }
 
   get storage() {
@@ -109,6 +128,25 @@ export interface MixinDeviceOptions<T> {
       this._deviceState = deviceManager.createDeviceState(this._deviceState.id, this._deviceState.setState);
     }
     this.mixinProviderNativeId = options.mixinProviderNativeId;
+
+    addScryptedInterfaceProperties({
+      thisObject: this,
+      getSet: (field: keyof DeviceState) => {
+        return (value: any) => {
+          this._lazyLoadDeviceState();
+          if (!this._deviceState)
+            console.warn('device state is unavailable. the device must be discovered with deviceManager.onDeviceDiscovered or deviceManager.onDevicesChanged before the state can be set.');
+          else
+            this._deviceState[field] = value;
+        }
+      },
+      getGet: (field: keyof DeviceState) => {
+        return () => {
+          this._lazyLoadDeviceState();
+          return this._deviceState?.[field];
+        }
+      }
+    })
   }
 
   get storage() {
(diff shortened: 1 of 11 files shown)


sdk/src/index.ts
```diff
@@ -6,13 +6,32 @@ import { DeviceBase, ScryptedInterfaceDescriptors, ScryptedInterfaceProperty, TY
...
+            this._deviceState[field] = value;
+        }
+      },
+      getGet: (field: keyof DeviceState) => {
+        return () => {
+          this._lazyLoadDeviceState();
+          return this._deviceState?.[field];
+        }
+      }
+    })
```
longzheng, line 34: I came up with this way to do this in a more type-safe manner, however it meant that we had to repeat it in both `class ScryptedDeviceBase` and `class MixinDeviceBase`

longzheng:
@koush There's a few places where `this.id` is called but it's optional so it might be undefined.

image

I'm not 100% sure if
- `DeviceBase` `id` should not be optional
- should assert `this.id` with a runtime check like `if (!this.id) throw new Error("device id not specified"`
- inline assert like `systemManager.getDeviceById(this.id!)?.nativeId`

sdk/src/storage-settings.ts
```diff
@@ -173,7 +173,11 @@ export class StorageSettings<T extends string> implements Settings {
             }
             return setting.defaultValue;
         };
-        const ret = parseValue(this.device.storage.getItem(key), setting, readDefaultValue, rawDevice);
+        const item = this.device.storage.getItem(key);
+        if (!item) {
```
koush, line 177: not found is valid, it will use the default value or return undefined. should not be throwing.

longzheng:
There's also a mismatch of types for `createDeviceState`

image

`this._deviceState.setState` is optional

```
export declare interface DeviceState {
  id?: string;
  setState?(property: string, value: any): Promise<void>;
}
```

but `createDeviceState` expects a `setState`

```
  createDeviceState?(id: string, setState: (property: string, value: any) => Promise<void>): DeviceState;
```

sdk/src/storage-settings.ts
longzheng: Right maybe in that case `parseValue` can be changed to accept a nullable `value`?

```ts
function parseValue(value: string, setting: StorageSetting, readDefaultValue: () => any, rawDevice?: boolean) {
```

becomes
```ts
function parseValue(value: string | null, setting: StorageSetting, readDefaultValue: () => any, rawDevice?: boolean) {
    if (value === null) {
        return readDefaultValue();
    }
    ...
```

longzheng: Or alternatively we can also inline in this function

```ts
        const item = this.device.storage.getItem(key);
        const ret = item !== null ? parseValue(item, setting, readDefaultValue, rawDevice) : readDefaultValue();
```

longzheng: @koush do you have a preference?

koush: parsing a value of null/undefined is valid, so the prior.

sdk/types/tsconfig.json
```diff
@@ -6,9 +6,10 @@
         "esModuleInterop": true,
         "sourceMap": true,
         "declaration": true,
-        "outDir": "dist"
+        "outDir": "dist",
+        "strict": true
     },
     "include": [
-        "gen/**/*"
+        "src/**/*"
```
koush, line 13: this should be gen

sdk/src/index.ts
```diff
@@ -31,7 +50,7 @@ export class ScryptedDeviceBase extends DeviceBase {
 
   get console() {
     if (!this._console) {
-      this._console = deviceManager.getDeviceConsole(this.nativeId);
+      this._console = deviceManager.getDeviceConsole?.(this.nativeId);
```
koush, line 53: this should never be nullish, and the signature on this method could be fixed.

sdk/types/tsconfig.json
longzheng: Ah yes, I was trying to fix the fact that `types.input.ts` was erroring because it wasn't being included by `tsconfig.json`.

I fixed this by creating a separate `tsconfig.json` in `src` and not emitting "noEmit": true`.

I added `tsc --project src` to `package.json build` so it is still type-checked.

https://github.com/koush/scrypted/pull/1308/commits/e595d0a55b7b9db45615a13b947feba4aeb2ebf9

sdk/src/index.ts
longzheng: Fixed

sdk/src/index.ts
```diff
@@ -115,7 +115,7 @@ export interface MixinDeviceOptions<T> {
   constructor(options: MixinDeviceOptions<T>) {
     super();
 
-    this.nativeId = systemManager.getDeviceById(this.id)?.nativeId;
+    this.nativeId = systemManager.getDeviceById(this.id!)?.nativeId;
```
longzheng, line 118: I'm manually asserting `this.id` is not undefined, but I'm not sure this should actually be changed in the "original types".

But the original types is a bit complicated because there's multiple layers

Firstly it's optional in `types.input.ts` https://github.com/koush/scrypted/blob/8fa0da7bd812c7badbc6c2364cfbe434e491e98a/sdk/types/src/types.input.ts#L30

Then your custom build script merges together all the properties and makes it optional anyway https://github.com/koush/scrypted/blob/0b6d61a801fbe559682e0ddd439d0f257583ba36/sdk/types/src/build.ts#L43

However I think we could change the build script to check the `isOptional` flag in `schema.json` and correctly write out `:` or `?:` depending on the actual property.

```json
		{
			"id": 2,
			"name": "ScryptedDevice",
			"kind": 256,
			"kindString": "Interface",
			"flags": {},
			"comment": {
				"summary": [
					{
						"kind": "text",
						"text": "All devices in Scrypted implement ScryptedDevice, which contains the id, name, and type. Add listeners to subscribe to events from that device."
					}
				]
			},
			"children": [
				{
					"id": 21,
					"name": "id",
					"kind": 1024,
					"kindString": "Property",
					"flags": {
						"isOptional": true
					},
```

sdk/src/index.ts
```diff
@@ -125,7 +125,7 @@ export interface MixinDeviceOptions<T> {
     // if the device state came from another node worker thread.
     // This should ultimately be removed at some point in the future.
     if ((this._deviceState as any).__rpcproxy_traps_all_properties && deviceManager.createDeviceState && typeof this._deviceState.id === 'string') {
-      this._deviceState = deviceManager.createDeviceState(this._deviceState.id, this._deviceState.setState);
+      this._deviceState = deviceManager.createDeviceState(this._deviceState.id, this._deviceState.setState!);
```
longzheng, line 128: Likewise I'm manually asserting this is non undefined, but I'm not sure if the original method should be non-nullable  

https://github.com/koush/scrypted/blob/8fa0da7bd812c7badbc6c2364cfbe434e491e98a/sdk/types/src/types.input.ts#L2281

sdk/types/package.json
```diff
@@ -7,10 +7,10 @@
   "license": "ISC",
   "scripts": {
     "prepublishOnly": "npm run build",
-    "build": "rimraf dist gen && typedoc && ts-node ./src/build.ts && tsc"
+    "build": "tsc --project src && rimraf dist gen && typedoc && ts-node ./src/build.ts && tsc"
```
koush, line 10: What's this?

sdk/src/index.ts
```diff
@@ -6,13 +6,32 @@ import { DeviceBase, ScryptedInterfaceDescriptors, ScryptedInterfaceProperty, TY
...
+  private _storage: Storage | undefined;
+  private _log: Logger | undefined;
+  private _console: Console | undefined;
+  private _deviceState: DeviceState | undefined;
 
   constructor(public readonly nativeId?: string) {
     super();
+
+    addScryptedInterfaceProperties({
+      thisObject: this,
```
koush, line 18: This a significant a behavioral change. It's running on every constructor and modifying the `this` object whereas prior it was modifying the class prototypes. While it may be valid, the change gives me the willies, and I think should mimic the existing behavior.

sdk/types/src/types.input.ts
```diff
@@ -935,7 +935,7 @@ export interface DeviceManager {
   /**
    * Get the console for the device given a native id.
    */
-  getDeviceConsole?(nativeId?: ScryptedNativeId): Console;
+  getDeviceConsole(nativeId?: ScryptedNativeId): Console;
```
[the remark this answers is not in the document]
longzheng: Thanks. I've removed all the optionals where they're not necessary.

sdk/src/index.ts
longzheng: Fair enough. I restored the original self executing function that modified the class prototypes. https://github.com/koush/scrypted/pull/1308/commits/576b59260d97e3e3ffb6cef149474693d60aece9

However to make TypeScript happy, I had to make the class `_deviceState` public because we're accessing it from outside the class.

https://github.com/koush/scrypted/pull/1308/commits/576b59260d97e3e3ffb6cef149474693d60aece9#diff-8db04fc25758a2af0f10ddcc8941574a24d2cfe1a3dc65f87204b1ca995145cbR12

This isn't perfect but I couldn't think of another solution.

koush: cast it as `any` and then grab it. 🤷🏽

koush: actually ya ts-ignore might work too

sdk/types/package.json
koush: ahh ok thanks

sdk/src/index.ts
longzheng: I tested unfortunately `ts-ignore` won't work for rename refactor either because TypeScript doesn't actually link the `_deviceState` to the property at all.

longzheng: If I have it as `public _deviceState` then the rename refactor works.

koush: yeah I think changing the signature back to `state: any` is best, the API needs to hide that.

longzheng: Although could argue `ts-ignore` is slightly better than `thisObject: any` since neither would allow for rename refactor but we get type-safety when calling `deviceBase._lazyLoadDeviceState();`

longzheng: I added ts-ignore, let me know if you disagree https://github.com/koush/scrypted/pull/1308/commits/907b4c2976a7b5bd4a5794386a9c52ab1c222ecf

sdk/src/index.ts
```diff
@@ -123,7 +123,7 @@ export interface MixinDeviceOptions<T> {
   get console() {
     if (!this._console) {
       if (deviceManager.getMixinConsole)
-        this._console = deviceManager.getMixinConsole(this.id, this.mixinProviderNativeId);
+        this._console = deviceManager.getMixinConsole(this.id!, this.mixinProviderNativeId);
```
koush, line 126: I just updated the sdk to make this type required, so can pull and remove this non-null assert.

sdk/src/index.ts
```diff
@@ -147,7 +147,7 @@ export interface MixinDeviceOptions<T> {
    * Fire an event for this device.
    */
   onDeviceEvent(eventInterface: string, eventData: any): Promise<void> {
-    return deviceManager.onMixinEvent(this.id, this, eventInterface, eventData);
+    return deviceManager.onMixinEvent(this.id!, this, eventInterface, eventData);
```
koush, line 150: and here too

sdk/src/index.ts
```diff
@@ -86,17 +86,17 @@ export interface MixinDeviceOptions<T> {
...
+  private _log: Logger | undefined;
+  private _console: Console | undefined;
   private _deviceState: WritableDeviceState;
   private _listeners = new Set<EventListenerRegister>();
 
   constructor(options: MixinDeviceOptions<T>) {
     super();
 
-    this.nativeId = systemManager.getDeviceById(this.id)?.nativeId;
+    this.nativeId = systemManager.getDeviceById(this.id!).nativeId;
```
koush, line 99: and here too

koush:
getting closer, I added some changes to accommodate some things your change pointed out.

longzheng:
I merge in fixes from https://github.com/koush/scrypted/pull/1321 so that should be merged first.

koush:
merged that, but conflicts on this for some reason

longzheng:
@koush resolved the conflicts

sdk/types/src/types.input.ts
```diff
@@ -1201,7 +1201,7 @@ export interface AmbientLightSensor {
   /**
    * The ambient light in lux.
    */
-  ambientLight: number;
+  ambientLight?: number;
```
longzheng, line 1204: @koush I noticed this type wasn't optional which seemed out of place, so I changed it.

sdk/src/index.ts
```diff
@@ -137,7 +137,7 @@ export interface MixinDeviceOptions<T> {
     });
   }
 
-  getMediaObjectConsole(mediaObject: MediaObject): Console {
+  getMediaObjectConsole(mediaObject: MediaObject): Console | undefined {
```
koush, line 140: I don't think this can ever return undefined. it will always return a console.

sdk/src/storage-settings.ts
```diff
@@ -173,7 +177,8 @@ export class StorageSettings<T extends string> implements Settings {
             }
             return setting.defaultValue;
         };
-        const ret = parseValue(this.device.storage.getItem(key), setting, readDefaultValue, rawDevice);
+        const item = this.device.storage.getItem(key);
```
koush, line 180: can revert this change to minimize diff, since the original throw you had earlier was removed.

sdk/types/src/types.input.ts
koush: Yep you're correct

sdk/src/index.ts
longzheng: Good point

sdk/src/storage-settings.ts
longzheng: Agreed

PR #1317, opened 2024-02-13
mqtt: add support for `ColorSettingTemperature` and `ColorSettingHsv`

nberardi:
I have added support for colored lights in the current MQTT design. In my testing this works with both Z2M and ZHA in Home Assistant.

### DIFF
--- a/plugins/mqtt/package.json
+++ b/plugins/mqtt/package.json
@@ -29,7 +29,6 @@
       ]
    },
    "dependencies": {
-      "@types/node": "^16.6.1",
       "aedes": "^0.46.1",
       "axios": "^0.23.0",
       "mqtt": "^4.2.8",
@@ -37,9 +36,10 @@
       "websocket-stream": "^5.5.2"
    },
    "devDependencies": {
-      "@scrypted/sdk": "file:../../sdk",
       "@scrypted/common": "file:../../common",
+      "@scrypted/sdk": "file:../../sdk",
+      "@types/node": "^18.4.2",
       "@types/nunjucks": "^3.2.0"
    },
-   "version": "0.0.76"
+   "version": "0.0.77"
 }
--- a/plugins/mqtt/tsconfig.json
+++ b/plugins/mqtt/tsconfig.json
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "module": "commonjs",
+        "module": "Node16",
         "target": "ES2021",
         "resolveJsonModule": true,
         "moduleResolution": "Node16",
--- a/plugins/mqtt/package-lock.json
+++ b/plugins/mqtt/package-lock.json
@@ -1,14 +1,13 @@
 {
    "name": "@scrypted/mqtt",
-   "version": "0.0.76",
+   "version": "0.0.77",
    "lockfileVersion": 2,
    "requires": true,
    "packages": {
       "": {
          "name": "@scrypted/mqtt",
-         "version": "0.0.76",
+         "version": "0.0.77",
          "dependencies": {
-            "@types/node": "^16.6.1",
             "aedes": "^0.46.1",
             "axios": "^0.23.0",
             "mqtt": "^4.2.8",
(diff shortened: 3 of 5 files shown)


plugins/mqtt/package.json
```diff
@@ -29,17 +29,18 @@
       ]
    },
    "dependencies": {
-      "@types/node": "^16.6.1",
       "aedes": "^0.46.1",
-      "axios": "^0.23.0",
+      "axios": "^1.6.7",
```
koush, line 33: this version of axios adds a 4MB dependency on mime db. it's why I haven't upgraded. I'm actually migrating off axios and using fetch directly since that is in node 18.

plugins/mqtt/package.json
```diff
@@ -29,17 +29,18 @@
...
+      "color-convert": "^2.0.1",
```
koush, line 34: I'd rather not introduce a dependency for a simple function here: from copilot: 

```ts
function rgbToHsv(r: number, g: number, b: number): [number, number, number] {
    r /= 255, g /= 255, b /= 255;

    let max = Math.max(r, g, b), min = Math.min(r, g, b);
    let h, s, v = max;

    let d = max - min;
    s = max == 0 ? 0 : d / max;

    if (max == min) {
        h = 0; // achromatic
    } else {
        switch (max) {
            case r: h = (g - b) / d + (g < b ? 6 : 0); break;
            case g: h = (b - r) / d + 2; break;
            case b: h = (r - g) / d + 4; break;
        }
        h /= 6;
    }

    return [h, s, v];
}
```

koush: this dependency looks relatively small and 1 level deep, so this could be fine. I generally prefer avoiding that on trivial stuff like this.

plugins/mqtt/src/util.ts
```diff
@@ -0,0 +1,227 @@
+import convert from 'color-convert'
```
koush, line 1: I think color-util.ts would be a better filename

nberardi:
I will downgrade it and resubmit. Sent from my iPhone, expect autocorrection issuesOn Feb 13, 2024, at 5:49 PM, Koushik Dutta ***@***.***> wrote:﻿
@koush commented on this pull request.



In plugins/mqtt/package.json:
>        "aedes": "^0.46.1",
-      "axios": "^0.23.0",
+      "axios": "^1.6.7",

this version of axios adds a 4MB dependency on mime db. it's why I haven't upgraded. I'm actually migrating off axios and using fetch directly since that is in node 18.

nberardi:
I was having the same debate with myself. I will remove and include
directly in ‘util.js’

plugins/mqtt/package.json
nberardi: I have completed this item

plugins/mqtt/package.json
nberardi: reverted

plugins/mqtt/src/util.ts
nberardi: done

nberardi:
I have addressed the above requests, and squashed them back into the commit

plugins/mqtt/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "module": "commonjs",
+        "module": "Node16",
```
koush, line 3: I dont think this is correct, you probably want this?

       "module": "CommonJS",
        "moduleResolution": "Node16",

nberardi: I will change if you like, but it usually helps me address odd Node issues during deployment. It is a valid module type. https://www.typescriptlang.org/tsconfig/#node16nodenext

nberardi: I will revert. It was likely troubleshooting that didn’t yield any change.

koush: I just noticed that my tsconfig is also complaining to change this, so I'm not sure what this is about. wasn't doing that before. might have been a node typings change. I'll investigate further later.

PR #1318, opened 2024-02-13
alexa: added support for `light`, `outlet`, and `fan` device types

nberardi:
Added support to the `alexa` plugin to support `Light`, `Outlet`, and `Fan` device types along with `ColorSettingsHsv` and `ColorSettingTemperature` beyond the standard interfaces.


nberardi:
You were a little too fast for me on this one, I forgot I had included `color-convert` but never used it.  It is now removed in #1319

PR #1321, opened 2024-02-15
SDK types build: calculate required/optional fields from schema

longzheng:
@koush I noticed you [added support for required properties](https://github.com/koush/scrypted/commit/bee77e121eafaa983b42e5c3634d74f5e3b8d43e) in the `build.ts` but I realised we can actually do this dynamically from the typedoc schema. 

In the future we can improve the types even further for `function toTypescriptType(type: any): string` and `function toPythonType(type: any): string {` but it was too much trouble so ignoring for now.

### DIFF
--- a/sdk/types/src/build.ts
+++ b/sdk/types/src/build.ts
@@ -2,12 +2,14 @@ import stringifyObject from 'stringify-object';
 import { ScryptedInterface, ScryptedInterfaceDescriptor } from "./types.input";
 import path from 'path';
 import fs from "fs";
+import packageJson from "../package.json"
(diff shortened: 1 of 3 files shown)


sdk/types/src/build.ts
```diff
@@ -2,12 +2,14 @@ import stringifyObject from 'stringify-object';
 import { ScryptedInterface, ScryptedInterfaceDescriptor } from "./types.input";
 import path from 'path';
 import fs from "fs";
+import packageJson from "../package.json"
```
longzheng, line 5: We can import `package.json` directly instead of using `JSON.parse(fs.readFileSync(`

PR #1323, opened 2024-02-17
CI build changed plugins

longzheng:
Add GitHub Actions workflow for any PRs that changes /plugins to build any plugins that changed

Fix snapshot test script


plugins/snapshot/test/test.ts
```diff
@@ -27,29 +26,4 @@ async function main1() {
     console.log('test1 done');
 }
 
-async function main2() {
-    const ret = await sharpFilterImage('/Users/koush/Downloads/151-1678381127261.jpg',
```
longzheng, line 31: Removed this since `sharpFilterImage` no longer exists

plugins/snapshot/test/test.ts
longzheng, line 1: I presume you used this file to play around with ffmpeg, so I left it

npm-install-base.sh
longzheng, line 1: I'm not very experienced with shell scripts so I'm not sure if this is the ideal way to split up this script

koush:
is this ready?

longzheng:
> is this ready?

Yes it is

PR #1333, opened 2024-02-19
Fix type of canMixin

longzheng:
Discovered while I'm fixing the strict types for snapshot plugin

I presume because the description specifically mentions "Returns null if a mixin can not be created" and the [usage of `canMixin` checks for null](https://github.com/koush/scrypted/blob/e5d60cf3e00e4bf399a75563aab9b85192af5224/common/src/autoenable-mixin-provider.ts#L51), that `null` is the correct response type here.


koush:
will this also allow undefined/void as well? those are valid falsy values I return elsewhere as well.

PR #1356, opened 2024-03-03
amcrest: Add support for Dahua VTO built-in access control lock as a `Lock`

rubyroobs:
I have a Dahua VTO video intercom that works great with the Amcrest plugin, but also has a built in access control door lock/unlock mechanism that I'd like to use in automation inside of scrypted. This adds support for locking/unlocking when explicitly enabled for Dahua Doorbells.

I have tested this in my system locally and believe it will work with most standard VTO configurations, but happy to submit a video or similar if needed to get this merged.

Thanks!


rubyroobs:
Updated the title to fit what seems to be the convention - please let me know if this needs changes otherwise feel free to merge :)

koush:
Looks good. Might be worthwhile to set this interface based on probing their APIs

rubyroobs:
> Looks good. Might be worthwhile to set this interface automatically during device creation based on probing their APIs

I intentionally kept this as something that can be toggled as the lock cannot be completely disabled on the VTO system itself. My system actually supports a "second lock" also which cannot be disabled (also annoyingly on the video intercom monitor indoors which causes confusion) but I couldn't think of a good way of making that work with the lock interface. 

Anyway, thank you for the swift review and merge! Hope this helps others too

PR #1360, opened 2024-03-05
server: add Python REPL support + introduce optional requirements.txt

bjia56:
Introduces `ptpython` as a remote REPL for Python plugins. Works with existing REPL plumbing.

Dependency tree of `ptpython`:
```
$ johnnydep --verbose=0 ptpython
name                               summary
---------------------------------  -------------------------------------------------------------------------------------------------
ptpython                           Python REPL build on top of prompt_toolkit
├── appdirs                        A small Python module for determining appropriate platform-specific dirs, e.g. a "user data dir".
├── jedi>=0.16.0                   An autocompletion tool for Python that can be used for text editors.
│   └── parso<0.9.0,>=0.8.3        A Python Parser
├── prompt-toolkit<3.1.0,>=3.0.34  Library for building powerful interactive command lines in Python
│   └── wcwidth                    Measures the displayed width of unicode strings in a terminal
└── pygments                       Pygments is a syntax highlighting package written in Python.
```


koush:
does python have optional dependencies? was thinking that maybe the pip install can be a separate requirements file that we can ignore the failure. so 2 pip installs.

bjia56:
There might be, I'd have to check. Though, I'm a bit hesitant to recommend splitting out into two requirements files since having one allows pip to resolve dependency constraints. If there are two, then I believe the second file will take precedence.

PR #1361, opened 2024-03-06
server: add workaround for asyncio to hook signals if gst runs on main thread

bjia56:
🐒 🔧


bjia56:
Most likely not needed after recent gst changes

PR #1380, opened 2024-03-15
add nscd to s6 docker

bjia56:


koush:
per conversation on discord this needs to be in a init script like the other daemons and run in foreground

PR #1393, opened 2024-03-27
Update install-scrypted-docker-compose.sh

kennethshsu:
I kept running into the error `docker: 'compose' is not a docker command.`. 

This should fix it.


koush:
update your docker per the installation docs. this will break everyone on a recent docker.

kennethshsu:
Yes, I follow the directions [here](https://github.com/koush/scrypted/wiki/Installation:-Docker-Compose-Linux). It installed docker-compose instead of docker v2 with compose as plugin.

PR #1440, opened 2024-04-22
rknn: initial rockchip object detector implementation

bjia56:
This currently only works on the RK3588(S) chip, e.g. Orange Pi 5 and similar. Also assumes installation is Docker, running in privileged mode.

### DIFF
--- a/plugins/rknn/src/rknn/plugin.py
+++ b/plugins/rknn/src/rknn/plugin.py
@@ -65,7 +65,7 @@ class RKNNPlugin(PredictPlugin):
             if ret != 0:
                 raise RuntimeError('Failed to load model: {}'.format(ret))
 
-            ret = rknn.init_runtime(core_mask=RKNNLite.NPU_CORE_ALL)
+            ret = rknn.init_runtime()
             if ret != 0:
                 raise RuntimeError('Failed to init runtime: {}'.format(ret))
 
--- a/plugins/rknn/src/rknn/plugin.py
+++ b/plugins/rknn/src/rknn/plugin.py
@@ -1,10 +1,6 @@
 import asyncio
-from concurrent.futures import Future
-from multiprocessing import shared_memory
 import os
 import platform
-import queue
-import threading
 from typing import Any, Coroutine, List, Tuple
 import urllib.request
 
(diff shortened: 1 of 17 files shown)


plugins/rknn/src/rknn/plugin.py
```diff
@@ -0,0 +1,119 @@
...
+    jobs: queue.Queue
+    executors: List[threading.Thread]
+
+    def __init__(self, nativeId=None):
+        super().__init__(nativeId)
+        ensure_compatibility()
+
+        if not os.path.exists(lib_path):
+            print('Downloading librknnrt.so from {}'.format(lib_download))
+            urllib.request.urlretrieve(lib_download, lib_path)
```
koush, line 79: us downloadFile? it handles versioning and caching.

plugins/rknn/src/rknn/plugin.py
```diff
@@ -0,0 +1,119 @@
...
+
+        model_path = self.downloadFile(model_download, os.path.basename(model_download))
+
+        self.jobs = queue.Queue()
+
+        self.executors = [
```
koush, line 85: and way to determine available cores rather than hard code?

plugins/rknn/src/rknn/plugin.py
```diff
@@ -0,0 +1,119 @@
...
+    def get_input_details(self) -> Tuple[int]:
+        return (IMG_SIZE[0], IMG_SIZE[1], 3)
+
+    def get_input_size(self) -> Tuple[int, int]:
+        return IMG_SIZE
+
+    async def detect_once(self, input: Image, settings: Any, src_size, cvss) -> Coroutine[Any, Any, Any]:
+        async def predict(input_tensor):
+            done_future = concurrent.futures.Future()
+            self.jobs.put((input_tensor, done_future))
```
koush, line 102: can use executors which provide an implicit queue, check the other plugins

plugins/rknn/src/rknn/plugin.py
bjia56: It seems that the rknn runtime library requires this to be in /usr/lib, I assume they are doing a dynamic runtime load with a full path using dlopen, since changing LD_LIBRARY_PATH to include the downloadFile directory does not work. Auditwheel (standard python tool to copy shared objects into the wheel) also ignores this library, so it is likely loaded as a runtime dependency

PR #1451, opened 2024-04-28
rknn: add more cpus

bjia56:
The additional CPUs are supported by Rockchip's model converter tool, but unknown if they will actually properly run Scrypted and the rknn plugin. These models have been converted and uploaded here: https://github.com/bjia56/scrypted-rknn/tree/main/models


bjia56:
The queue-based runtime selection approach was reverted because it caused a drop of 5-10 detections per second compared to the dict method.

koush:
Weird, need to switch tf over to that as well. May be a python lock perf issue.

Issue #1453, opened 2024-04-29
Incompatible with Windows Node 20..12.2


longzheng:
**Describe the bug**

I know technically [Scrypted's Windows install script](https://github.com/koush/scrypted/blob/main/install/local/install-scrypted-dependencies-win.ps1#L11) hard-codes the Node version to 20.11.1 so any higher version is unsupported, but I thought I'd document this bug with Node v20.12.2 in case anyone manually updates their Node.js version like me.

I noticed that the Scrypted service was not starting.

I tried running the service manually with `node %USERPROFILE%\.scrypted\service.js` and got the following exception

```
Microsoft Windows [Version 10.0.22631.3447]
(c) Microsoft Corporation. All rights reserved.

C:\Users\Long>node %USERPROFILE%\.scrypted\service.js
node:internal/child_process:421
    throw new ErrnoException(err, 'spawn');
    ^

Error: spawn EINVAL
    at ChildProcess.spawn (node:internal/child_process:421:11)
    at Object.spawn (node:child_process:761:9)
    at Object.<anonymous> (C:\Users\Long\.scrypted\service.js:8:15)
    at Module._compile (node:internal/modules/cjs/loader:1369:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1427:10)
    at Module.load (node:internal/modules/cjs/loader:1206:32)
    at Module._load (node:internal/modules/cjs/loader:1022:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12)
    at node:internal/main/run_main_module:28:49 {
  errno: -4071,
  code: 'EINVAL',
  syscall: 'spawn'
}

Node.js v20.12.2
```

After a quick Google, I found other people [documenting the same error code](https://github.com/nodejs/node/issues/52554) which is the result of a [security behaviour change in 20.12.2](https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2) specifically for Windows.

> It is important to note that there has been a breaking change for Windows users who utilize child_process.spawn and child_process.spawnSync. Node.js will now error with EINVAL if a .bat or .cmd file is passed to child_process.spawn and child_process.spawnSync without the shell option set. If the input to spawn/spawnSync is sanitized, users can now pass { shell: true } as an option to prevent the occurrence of EINVALs errors. 

Of course Scrypted's `service.js` tries to launch `npx.cmd` which is affected.

I tried modifying `service.js` to include `shell: true`

```js
[...]

PR #1455, opened 2024-04-30
Fix Node 20.12.2 child_process.spawn .cmd EINVAL on Windows

longzheng:
Fixes https://github.com/koush/scrypted/issues/1453 incompatibility with Node 20.12.2 on Windows. https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2

I looked for all usages of `child_process.spawn` which invoked a `.cmd` file and added `shell: true` (conditionally since I didn't want to change non-Windows behavior)

I also wrapped the `NPX_PATH_ESCAPED` because the path will contain spaces e.g. `C:\\Program Files\\nodejs\\npx.cmd` which doesn't work with `shell: true`

There were other usages of `child_process.spawn` which were for `ffmpeg` and `ffplay` which I don't think needs any change since those are `.exe` files.

### DIFF
--- a/install/local/install-scrypted-dependencies-win.ps1
+++ b/install/local/install-scrypted-dependencies-win.ps1
@@ -45,7 +45,7 @@ try {
 catch (e) {
 }
 const child_process = require('child_process');
-child_process.spawn('$($NPX_PATH_ESCAPED)', ['-y', 'scrypted', 'serve'], {
+child_process.spawn('$NPX_PATH_ESCAPED', ['-y', 'scrypted', 'serve'], {
     stdio: 'inherit',
     // allow spawning .cmd https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2
     shell: true,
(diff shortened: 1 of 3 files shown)


install/local/install-scrypted-dependencies-win.ps1
```diff
@@ -34,7 +34,8 @@ $SCRYPTED_HOME_ESCAPED_PATH = $SCRYPTED_HOME.replace('\', '\\')
 npm install --prefix $SCRYPTED_HOME @koush/node-windows --save
 
 $NPX_PATH = (Get-Command npx).Path
-$NPX_PATH_ESCAPED = $NPX_PATH.replace('\', '\\')
+# The path needs double quotes to handle spaces in the directory path
+$NPX_PATH_ESCAPED = `"${$NPX_PATH.replace('\', '\\')}"`
```
longzheng, line 38: @koush I couldn't quite figure out why you changed the NPX path to be the full path in https://github.com/koush/scrypted/commit/ddb6a3a47d68503090dd781522303a87869575a1 as opposed to just `npx.cmd`, but I guess you had a good reason.

longzheng: I'm guessing because it's running as a service it needs the full path since it doesn't have the npm directory in the `Path` environment variable?

PR #1461, opened 2024-05-05
homekit: fix late 2way setup

xdissent:
When my ring doorbell sets up a streaming session, the first message on the audio return channel isn't a request to setup 2way. In that case, the "once" handler would fire (and bail) and when the real 2way setup message would arrive, the setup listener was no longer registered and nothing would happen. This PR moves the 2way setup into the always-on listener, and creates a guard to prevent multiple setup requests from running concurrently. It should also allow retrying the setup if something temporarily went wrong on a previous attempt.

### DIFF
--- a/plugins/homekit/src/types/camera/camera-streaming.ts
+++ b/plugins/homekit/src/types/camera/camera-streaming.ts
@@ -354,12 +354,11 @@ export function createCameraStreamingDelegate(device: ScryptedDevice & VideoCame
             if (twoWayAudio) {
                 let rtspServer: RtspServer;
                 let track: string;
-                let playing = false;
-                let initializing = false;
+                let twoWayAudioState: 'stopped' | 'starting' | 'started' = 'stopped';
 
-                const initialize = async () => {
+                const start = async () => {
                     try {
-                        initializing = true;
+                        twoWayAudioState = 'starting';
                         const { clientPromise, url } = await listenZeroSingleClient();
                         const rtspUrl = url.replace('tcp', 'rtsp');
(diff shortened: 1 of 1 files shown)


plugins/homekit/src/types/camera/camera-streaming.ts
```diff
@@ -355,14 +355,11 @@ export function createCameraStreamingDelegate(device: ScryptedDevice & VideoCame
...
                 let track: string;
                 let playing = false;
-                session.audioReturn.once('message', async buffer => {
-                    try {
-                        const decrypted = srtpSession.decrypt(buffer);
-                        const rtp = RtpPacket.deSerialize(decrypted);
-
-                        if (rtp.header.payloadType !== session.startRequest.audio.pt)
-                            return;
+                let initializing = false;
```
koush, line 358: This code seems correct, but it was a bit confusing to read due to the addition of a state. Can you make a change with something like:

```ts
let twoWayAudioState = 'stopped' | 'starting' | 'started' = 'stopped';
```

Then use this variable throughout to manage check change/check.

koush: Related: this would seem to allow restarting of the two way audio if it terminates for some reason. Incidentally, I may consider terminating two way audio automatically if homekit stops sending packets (which it does when the mic is disabled on the ios home app). Many cameras operate with echo cancellation during two way audio, which mutes incoming audio to some degree.

xdissent: Done. Much easier to read, I agree. And yes, it does allow restarting of the two way audio!

PR #1472, opened 2024-05-11
tapo: normalize fetch header case

xdissent:
Fixes two-way audio on tapo cameras that require case-sensitive http headers.


koush:
I think respecting the caller's initial casing is the better fix here:

https://github.com/koush/scrypted/commit/a807766699a5245ea7f550a1e6963b1163705634

xdissent:
Ah, yeah that could work but needs to be done in `authHttpFetch` too. That creates a Headers immediately and then calls `httpFetch()` with just the Headers obj.

PR #1477, opened 2024-05-17
server: patch os.networkInterfaces for android compatibility

bjia56:
Replace `os.networkInterfaces` with a function that reads from an environment variable, if the install target is set to `android`. This patch is done early in the node process stack for both scrypted server and node plugins.

Undos #1474 since the workaround is no longer needed after this change.

### DIFF
--- a/server/src/server-settings.ts
+++ b/server/src/server-settings.ts
@@ -29,14 +29,7 @@ function nodeIpAddress(family: number): string[] {
         ...costlyNetworks,
     ];
 
-    let interfaces: any;
-    try {
-        interfaces = os.networkInterfaces();
-    } catch {
-        // bjia56: When running in secured environments like UserLAnd in Android, os.networkInterfaces()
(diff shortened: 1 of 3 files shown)


koush:
I think this would be better patched in the node runtime itself, or however you are launching it. So it happens automatically for all plugins.

server/src/server-settings.ts
```diff
@@ -29,14 +29,7 @@ function nodeIpAddress(family: number): string[] {
         ...costlyNetworks,
     ];
 
-    let interfaces: any;
```
koush, line 32: Feel like this change was q mistake. I don't think the server should be responsible for managing quirks like this.

koush:
I think the best solution is to `require` some js file in the volume which manages this kinda stuff. Revert the previous change.

koush:
closing per our discussion on requiring generic combat file (via some process.env variable?) on startup.

PR #1487, opened 2024-05-28
Add security_opt to allow the container to talk to host avahi daemon

jonathanyip:
While running the Avahi daemon on the host and scrypted in a docker container, I noticed that the homekit plugin would fail with the following exception:
Screenshot 2024-05-27 at 11 20 18 PM

Debugging the homebridge lib, I found out that it was caused by this error:
```
'An AppArmor policy prevents this sender from sending this message to this recipient; type="method_call", sender="(null)" (inactive) interface="org.freedesktop.DBus" member="Hello" error name="(unset)" requested_reply="0" destination="org.freedesktop.DBus" (bus)'
```

I found this snippet in the homebridge docs (https://github.com/homebridge/homebridge/wiki/mDNS-Options#avahi-linux-and-docker)
> Unless you've disabled AppArmor on your Linux distro, or it's disabled by default, you'll also need to set a security_opt setting of apparmor:unconfined, otherwise AppArmor will block the container's attempt to talk to Avahi via dbus

After adding those lines to my own docker-compose file and restarting, it seemed to work.

This PR adds those lines to the docker-compose.yml and installation script, when the user wants to use host avahi

### DIFF
--- a/install/docker/docker-compose.yml
+++ b/install/docker/docker-compose.yml
@@ -35,7 +35,7 @@ services:
             # Avahi can be used for network discovery by passing in the host daemon
             # or running the daemon inside the container. Choose one or the other.
             # Uncomment next line to run avahi-daemon inside the container.
-            # See volumes section below to use the host daemon.
+            # See volumes and security_opt section below to use the host daemon.
             # - SCRYPTED_DOCKER_AVAHI=true
 
             # NVIDIA (Part 1 of 4)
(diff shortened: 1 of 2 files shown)


koush:
Thanks for the find. Honestly this should likely be the default. I had to remove avahi as the default option recently, because it was failing for a lot of people. I suspect this was the issue: it was mounted correctly but communicating with the socket was not allowed.

PR #1493, opened 2024-06-03
Enable TypeScript strict for packages/auth-fetch

longzheng:
Adding `tsconfig.json` `strict: true` to `packages/auth-fetch`

I had to tweak the types of `server/src/fetch/index.ts` and `server/src/fetch/http-fetch.ts` since `packages/auth-fetch/src/auth-fetch.ts` was calling functions with `undefined` values.


server/src/fetch/http-fetch.ts
```diff
@@ -41,19 +41,21 @@ const StreamParser: FetchParser<IncomingMessage> = {
...
     switch (responseType) {
         case 'json':
             return JSONParser;
         case 'text':
             return TextParser;
         case 'readable':
             return StreamParser;
+        case 'buffer':
+        case undefined:
+            return BufferParser;
```
longzheng, line 54: I find explicitly specifying every case in `switch` syntax more obvious than relying on fall through.

koush: This should be default.

Or throw below the switch. This code fails for null value now.

koush: Would actually prefer not to have this specific change. For the scope of strict typing updates, I don't think there should be any potential behavioral changes (ie null case failure here). Same as for the other switch below.

longzheng: That's fair.

Although in the future if everything is updated to `strict: true` then it will be safe to assume either `null` will never be passed in as a value or there will be a type error if `null` is passed in.

PR #1494, opened 2024-06-03
Enable TypeScript strict for packages/cli

longzheng:
Adding `tsconfig.json` `strict: true` to `packages/cli`

### DIFF
--- a/packages/cli/src/service.ts
+++ b/packages/cli/src/service.ts
@@ -90,7 +90,13 @@ export async function installServe(installVersion: string, ignoreError?: boolean
     const installJson = path.join(installDir, 'install.json');
     try {
         const { version } = JSON.parse(fs.readFileSync(installJson).toString());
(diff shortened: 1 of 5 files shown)


packages/cli/src/service.ts
```diff
@@ -90,7 +90,13 @@ export async function installServe(installVersion: string, ignoreError?: boolean
     const installJson = path.join(installDir, 'install.json');
     try {
         const { version } = JSON.parse(fs.readFileSync(installJson).toString());
-        if (semver.parse(process.version).major !== semver.parse(version).major)
+        const processSemver = semver.parse(process.version);
+        if (!processSemver)
+            throw new Error('error parsing process version');
+        const installSemver = semver.parse(version);
+        if (!installSemver)
+            throw new Error('error parsing install.json version');
```
longzheng, line 98: This is just guarding against `semver.parse` returning `null` if it fails to parse

PR #1497, opened 2024-06-15
reolink: save GetAiState to hasObjectDetector so object detector can work

gtalusan:
I purchased some CX410's the other day and noticed that the Video Analysis plugin wouldn't recognize the cameras as it does my POE Doorbell.

With this patch, I can create a new smart motion sensor for the camera, and the object detector drop down will show 'dog_cat', 'people', and 'vehicle' as expected.

### DIFF
--- a/plugins/reolink/src/main.ts
+++ b/plugins/reolink/src/main.ts
@@ -91,7 +91,7 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, Reboot, Intercom,
         }
 
         try {
-            const ai: AIState = this.storageSettings.values.hasObjectDetector[0]?.value;
+            const ai: AIState = this.storageSettings.values.hasObjectDetector?.value;
             const classes: string[] = [];
 
             for (const key of Object.keys(ai)) {
@@ -203,6 +203,8 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, Reboot, Intercom,
 
                     if (!classes.length)
                         return;
+
+                    this.storageSettings.values.hasObjectDetector = ai;
 
                     hasSucceeded = true;
                     const od: ObjectsDetected = {
--- a/plugins/reolink/src/reolink-api.ts
+++ b/plugins/reolink/src/reolink-api.ts
@@ -122,7 +122,7 @@ export class ReolinkCameraClient {
             responseType: 'json',
         });
         return {
-            value: response.body?.[0]?.value as AIState,
+            value: (response.body?.[0]?.value || response.body?.value) as AIState,
             data: response.body,
         };
     }


koush:
I think the intent was for the ai state has object types available, I think this change makes empty results also valid.

getAiState payload:

```json
[
   {
      "cmd" : "GetAiState",
      "code" : 0,
      "value" : {
         "channel" : 0,
         "dog_cat" : {
            "alarm_state" : 0,
            "support" : 0
         },
         "face" : {
            "alarm_state" : 0,
            "support" : 0
         },
         "people" : {
            "alarm_state" : 0,
            "support" : 1
         },
         "vehicle" : {
            "alarm_state" : 0,
            "support" : 0
         }
      }
   }
]
```

plugins/reolink/src/main.ts
```diff
@@ -91,7 +91,7 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, Reboot, Intercom,
         }
 
         try {
-            const ai: AIState = this.storageSettings.values.hasObjectDetector[0]?.value;
+            const ai: AIState = this.storageSettings.values.hasObjectDetector?.value;
```
koush, line 94: this change looks overly permissive. need to see the camera payload.

gtalusan: This is the payload coming out of the camera.
```
{
  value: {
    channel: 0,
    dog_cat: { alarm_state: 0, support: 1 },
    face: { alarm_state: 0, support: 0 },
    people: { alarm_state: 0, support: 1 },
    vehicle: { alarm_state: 0, support: 1 }
  },
  data: [ { cmd: 'GetAiState', code: 0, value: [Object] } ]
}
```

plugins/reolink/src/main.ts
```diff
@@ -204,6 +204,8 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, Reboot, Intercom,
                     if (!classes.length)
                         return;
 
+                    this.storageSettings.values.hasObjectDetector = ai;
```
gtalusan, line 207: @koush regarding the empty response or if support = 0.. the code prior to this line will only fill up `classes` if support is 1.

plugins/reolink/src/main.ts
koush: This is a different payload, as it is not enclosed by a JSON array. Change needs to be made to check both the original doorbell payload as well as this new payload (removing the [0] index).

koush:
now that I think about it, the doorbell doesn't even use this code path anymore as the events are delivered via onvif. does this camera support onvif events too? that may be preferable since it won't need to poll.

koush:
This may be the necessary fix for getAiState:

```ts
            value: (response.body?.[0]?.value || response.body?.value) as AIState,

File: plugins/reolink/src/main.ts (hunk 1 above)
Comment (gtalusan) [reply plugins/reolink/src/main.ts:94]: I don't see a JSON array coming back for either my POE doorbell or these cameras.

Was this originally coded against an NVR?

**Comment by gtalusan** [comment]
Unsure at the moment if this camera supports ONVIF events.  I can investigate afterward.

Forced push an update to include:

```
            value: (response.body?.[0]?.value || response.body?.value) as AIState,
```

File: plugins/reolink/src/main.ts [hunk 3]
```diff
@@ -91,7 +91,7 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, Reboot, Intercom,
         }
 
         try {
-            const ai: AIState = this.storageSettings.values.hasObjectDetector[0]?.value;
```
koush, line 94: I believe the plain hasObjectDetector can be checked now? I don't think the value property exists since its destructured in getAiState

koush:
Ah yeah you're correct. The data is only used for debugging in the console.

So my understanding is that this getAiState is currently never saved, and that's also part of the fix? I think I must have been doing that at some point, but when I switched the doorbell over to onvif that old code got deleted.

gtalusan:
Yup, exactly!

PR #1506, opened 2024-06-23
reolink: add siren support

gtalusan:
This changes the camera to be a DeviceProvider, so that a Siren device can be provided by the camera if siren mode is supported.  Siren support is inferred via the GetAbility API.

For my PoE doorbell, I noticed that the alarm_mode = "manul" won't trigger the siren.  Using the siren's timed mode instead and pumping the API is a workaround.


plugins/reolink/src/main.ts
```diff
@@ -521,6 +596,7 @@ class ReolinkProvider extends RtspProvider {
             ScryptedInterface.Camera,
             ScryptedInterface.AudioSensor,
             ScryptedInterface.MotionSensor,
+            ScryptedInterface.DeviceProvider,
```
koush, line 599: don't add this interface unless it has a siren.

gtalusan: Got it.  Updated to push the DeviceProvider interface when the siren ability is determined.

PR #1508, opened 2024-06-26
reolink: fix deviceprovider for cameras with siren

gtalusan:
Looks like I inadvertently caused a bug.  It's more than likely I had a cached response in my storage settings from intermediate code.

Anyway, my "production" instance of Scrypted synced up with the latest publish and I couldn't see the sirens after adding in my test camera.

This fixes it.

Apologies as I'm still learning my way around this codebase.


koush:
Yeah I put in a couple other fixes, device provider was always being added (despite the change to fix that). setting the ai state was causing a settings refresh once per second, need to be more discerning about doing that. can check the commit history on that path.

PR #1509, opened 2024-06-27
reolink: use Login json api to get around URL escaping limitations with some firmware

gtalusan:
I noticed that I was unable to login to some of my cameras with Scrypted because my passwords have hyphens and spaces in them.  In fact, the latest firmware available for my RLN36 NVR and CX410s requires the passwords to have passwords with mixed alpha-numeric and various symbols as can be seen from this screenshot:

Screenshot 2024-06-26 at 22 06 11

Here we'll switch over to using the Login API which has a JSON payload for the password.  Each of the functions will check if the token has expired, and if so, we will re-perform the login to get a new token.

### DIFF
--- a/plugins/reolink/src/main.ts
+++ b/plugins/reolink/src/main.ts
@@ -413,6 +413,23 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
         return ret;
     }
 
+    addRtspCredentials(rtspUrl: string) {
+        const url = new URL(rtspUrl);
+        if (url.protocol !== 'rtmp:') {
+            url.username = this.storage.getItem('username');
+            url.password = this.storage.getItem('password') || '';
+        } else {
+            const params = url.searchParams;
+            params.set('token', this.client.token);
+        }
+        return url.toString();
+    }
+
+    async createVideoStream(vso: UrlMediaStreamOptions): Promise<MediaObject> {
+        await this.client.login();
+        return super.createVideoStream(vso);
+    }
+
     async getConstructedVideoStreamOptions(): Promise<UrlMediaStreamOptions[]> {
         this.videoStreamOptions ||= this.getConstructedVideoStreamOptionsInternal().catch(e => {
             this.constructedVideoStreamOptions = undefined;
@@ -500,8 +517,6 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
                 const params = streamUrl.searchParams;
                 params.set("channel", this.getRtspChannel().toString())
                 params.set("stream", '0')
-                params.set("user", this.getUsername())
-                params.set("password", this.getPassword())
                 stream.url = streamUrl.toString();
                 stream.name = `RTMP ${stream.id}`;
             } else if (stream.container === 'rtsp') {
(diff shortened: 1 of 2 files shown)


koush:
does this login api exist on all reolink cameras

gtalusan:
Yes, it appears in Version 1 2016-11 of their API documentation.

gtalusan:
So to be clear, the code prior to the recent `!skipValidation` check worked for me.  Switching to the JSON API should make it work for cameras stuck on older or unmaintained firmware.

koush:
understood. this change would be better if you created a new method `requestWithLogin` and call login then request inside that. then change all the API calls to requestWithLogin.

gtalusan:
Ok, updated.  I had to tighten up `urlOrOptions` to just be a bag of options and forced the code to use `url` as an actual `URL` object instead of a string, however the awkwardness is only within `requestWithLogin`.

koush:
I just remembered that the simple password requirement extends to RTMP as well. The login code here is only half the story.
I think rolling back my change will start causing issues (again) with devices that can't accept complex passwords via RTMP. What reolink devices do you have on hand? I only have the doorbell.
Incidentally I tried changing the doorbell to a complex password in their web admin... afterwards I couldn't log into the web admin anymore. Had to reset it.

koush:
getConstructedVideoStreamOptions should return fairly quickly. I actually don't think the token needs to be in the url returned there. the url doesn't need to be returned at all actually. getVideoStream should refresh the token.

gtalusan:
Ok I'll dig a bit further.

koush:
I think this function needs to be overridden to add the token:

https://github.com/koush/scrypted/blob/ea606de22f013a6c716aaf22aba3884cf08673ef/plugins/rtsp/src/rtsp.ts#L84

gtalusan:
Ok, I overrode `createVideoStream` and `addRtspCredentials` to inject the token for rtmp:// URLs.

RTSP doesn't appear to support `&token=` in my testing.

koush:
> Ok, I overrode `createVideoStream` and `addRtspCredentials` to inject the token for rtmp:// URLs.
> 
> RTSP doesn't appear to support `&token=` in my testing.

RTSP likely requires the username and token as credentials and not parameters.

ie rtsp://username:token@192.168.2.222/....

gtalusan:
Unfortunately not.  Got a 401 Unauthorized.

plugins/reolink/src/probe.ts
```diff
@@ -1,25 +0,0 @@
-import { AuthFetchCredentialState, authHttpFetch } from '@scrypted/common/src/http-auth-fetch';
-import https from 'https';
-
-export const reolinkHttpsAgent = new https.Agent({
-    rejectUnauthorized: false,
-});
-
-export async function getMotionState(credential: AuthFetchCredentialState, username: string, password: string, address: string, channelId: number) {
```
koush, line 8: please restore this file, it's used elsewhere.

gtalusan: Ok updated

koush:
this change broke doorbell snapshots. reverting fixed it.

gtalusan:
Interesting.  Works here.  Do you have any debug logs?

koush:
it actually seems to be working now, after I reapplied the change. I was getting a json error before.

koush:
reported by another user as well. I switched to using username/password if it works, and falling back otherwise.

PR #1533, opened 2024-07-21
rebroadcast: support env vars + upstream encoder args

bjia56:


plugins/prebuffer-mixin/src/main.ts
```diff
@@ -591,6 +591,9 @@ class PrebufferSession {
         //   ffmpegInput.inputArguments.unshift('-use_wallclock_as_timestamps', '1');
 
         // extraOutputArguments must contain full codec information
+        if (ffmpegInput.h264EncoderArguments) {
```
koush, line 594: move this up a line to preserve comment meant for below.

bjia56: done

plugins/prebuffer-mixin/src/ffmpeg-rebroadcast.ts
```diff
@@ -134,6 +134,10 @@ export async function startParserSession<T extends string>(ffmpegInput: FFmpegIn
 
 
     const args = ffmpegInput.inputArguments.slice();
+    const env = {
```
koush, line 137: I mean this should only be set if ffmpegInput.env is used, otherwise use undefined, which is the existing behavior.

koush: env = ffmpegInput.env ? { ...process.env, ...ffmpegInput.env } : undefined;

PR #1540, opened 2024-08-03
sdk: add TTYSettings

bjia56:
Intended to be used by btop plugin to add the path where btop is installed to the interactive terminal.


bjia56:
Another possible implementation is to expose only one function, `getTerminalSettings`, and have that return a list (similar to the existing `getSettings`), where some list items may be labeled as "PATH"

koush:
should return something like

```ts
getTerminalSettings(): {
  paths?: string[];
}
```

koush:
Should maybe all be called TTYSettings per the existing TTY interface?

PR #1542, opened 2024-08-05
server: allow plugins to customize python installs

bjia56:
image


bjia56:
I think it would be nice to merge the options into `pythonVersion` although not sure how best to do that given the existing platform/arch-specific keying inside that.

bjia56:
Updated structure
image

PR #1546, opened 2024-08-07
tempest plugin

DMontgomery40:
early stage but somewhat functional plugin for tempest weather station


koush:
this will not go into the mono repo, only core/camera plugins go in there. can be forked to scryptedapp as an official plugin or self published by you. whatever you prefer.

DMontgomery40:
ahh my bad, self published for now, if it becomes real can i get that sweet sweet dev flair in discord lol?

Issue #1520, opened 2024-07-07
The Scrypted plugin for OpenCV fails to start, throwing an ImportError for the `Image` module from the `PIL` library.


cwawak:
**Describe the bug**
The Scrypted plugin for OpenCV fails to start, throwing an ImportError for the `Image` module from the `PIL` library.

**To Reproduce**
1. Start the Scrypted server.
2. Load the OpenCV plugin.
3. Observe the error in the logs.

**Expected behavior**
The OpenCV plugin should start without errors, and the `Image` module from `PIL` should be successfully imported.

**Screenshots**
N/A

**Server (please complete the following information):**
- OS: Fedora 40
- Installation Method: Podman
- Image: `docker.io/koush/scrypted:v0.114.0-jammy-full`

**Hardware Model (please complete the following information):**
- Device: x86_64 server
  
**Client (please complete the following information, if applicable):**
- Software: N/A

**Additional context**
This issue is running under Fedora using Podman, with the following image:
`docker.io/koush/scrypted:v0.114.0-jammy-full`

Traceback:
```
linux x64 #1 SMP PREEMPT_DYNAMIC Thu May  2 18:50:49 UTC 2024
server version: 0.114.0
plugin version: @scrypted/opencv 0.0.91
full
########################
7/7/2024, 12:55:07 PM
########################
python version: python3.10
interpreter: /usr/bin/python3
pip target: /server/volume/plugins/@scrypted/opencv/python3.10-Linux-x86_64-20240317
requirements.txt (up to date)
# must ensure numpy is pinned to prevent dependencies with an unpinned numpy from pulling numpy>=2.0.
numpy==1.26.4
imutils>=0.5.0
opencv-python==4.10.0.82
Pillow==10.3.0

plugin start/fork failed
Traceback (most recent call last):
  File "/server/node_modules/@scrypted/server/python/plugin_remote.py", line 379, in loadZip
    return await self.loadZipWrapped(packageJson, getZip, options)
  File "/server/node_modules/@scrypted/server/python/plugin_remote.py", line 690, in loadZipWrapped
    from main import create_scrypted_plugin  # type: ignore
  File "/server/volume/plugins/@scrypted/opencv/zip/1-e7219c81a6fa17e84dfbc7760e7a2a2a.zip/main.py", line 1, in <module>
    from opencv import OpenCVPlugin
[...]

PR #1553, opened 2024-08-14
server: add unzipped path to python module lookup path

bjia56:
So... I'm not sure how this ever worked.

The unzipped path wasn't added to the Python system path, so it was never explicitly available when looking up modules. It seems that pure-Python sources were unaffected, however loading binary extension modules bundled together inside the plugin (e.g. plugins compiled by [Nuitka ](https://github.com/Nuitka/Nuitka)) would never get found. This change adds the unzipped path but keeps existing behavior

This change will need some additional beta testing before release. I am able to load existing plugins just fine (OpenCV, Python Codecs, OpenVINO, btop, btop-camera).

### DIFF
--- a/server/python/plugin_remote.py
+++ b/server/python/plugin_remote.py
@@ -798,9 +798,7 @@ class PluginRemote:
                 print("requirements.txt (up to date)")
                 print(str_requirements)
 
-            sys.path.insert(0, zipPath)
-            if not debug:
-                sys.path.insert(0, plugin_zip_paths.get("unzipped_path"))
+            sys.path.insert(0, plugin_zip_paths.get("unzipped_path"))
             sys.path.insert(0, pip_target)
 
         self.systemManager = SystemManager(self.api, self.systemState)
--- a/server/python/plugin_remote.py
+++ b/server/python/plugin_remote.py
@@ -799,6 +799,8 @@ class PluginRemote:
                 print(str_requirements)
 
             sys.path.insert(0, zipPath)
+            if not debug:
+                sys.path.insert(0, plugin_zip_paths.get("unzipped_path"))
             sys.path.insert(0, pip_target)
 
         self.systemManager = SystemManager(self.api, self.systemState)


bjia56:
My compiled plugin file structure looks like this:
image

With `main.py`:
```python
from plugin import MyPlugin

def create_scrypted_plugin():
    return MyPlugin()
```

koush:
could be related to this https://github.com/koush/scrypted/issues/1520#issuecomment-2287703749

server/python/plugin_remote.py
```diff
@@ -799,6 +799,8 @@ def read_requirements(filename: str) -> str:
                 print(str_requirements)
 
             sys.path.insert(0, zipPath)
+            if not debug:
```
koush, line 802: why is the debug check necessary?

koush: should maybe also remove the zipPath altogether if using unzippedPath?

bjia56: There is a comment higher up in the function that says the debugger wanted a stable path, so zipPath points to the zip file. I am not sure what the technical reasons are for that, so I kept the behavior unmodified for the debug route

koush: I think that requirement is no longer necessary as I think the vscode debugger now supports relative paths when resolving remote/local roots. I'll play with this and see.

bjia56: Got it, I'll push up a version that removes it

koush:
hm this remote root thing isn't working properly, and it seems to be a vscode issue
while the breakpoints do work, it opens a readonly main.py rather than the one that is in the local path. makes debugging a pain.

koush:
I'm trying your original change with the debug check... I assume this works for you because the code isn't obfuscated while debugging?

bjia56:
I haven't used the debugger in a while now but since I have the source python files before compilation, I can always just bundle up the originals when debugging

PR #1558, opened 2024-08-17
webrtc: convert RTCSignalingChannel to FFmpegInput + consolidate all converters

bjia56:
I was able to test this by converting a WebRTC mixin device (since that implements RTCSignalingChannel) to FFmpegInput, then ffprobe-ing the returned RTSP.

### DIFF
--- a/plugins/webrtc/src/main.ts
+++ b/plugins/webrtc/src/main.ts
@@ -13,7 +13,7 @@ import ip from 'ip';
 import net from 'net';
 import os from 'os';
 import { DataChannelDebouncer } from './datachannel-debouncer';
-import { RTC_BRIDGE_NATIVE_ID, WebRTCConnectionManagement, createRTCPeerConnectionSink, createTrackForwarder } from "./ffmpeg-to-wrtc";
+import { WebRTCConnectionManagement, createRTCPeerConnectionSink, createTrackForwarder } from "./ffmpeg-to-wrtc";
 import { stunServers, turnServers, weriftStunServers, weriftTurnServers } from './ice-servers';
 import { waitClosed } from './peerconnection-util';
 import { WebRTCCamera } from "./webrtc-camera";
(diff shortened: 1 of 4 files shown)


koush:
the multiple buffer converters thing is getting unwieldy, this was added a while ago

image

bjia56:
That looks perfect, I can see about porting the old RTC bridge over

plugins/webrtc/src/main.ts
```diff
@@ -772,4 +784,41 @@ class WebRTCBridge extends ScryptedDeviceBase implements BufferConverter {
...
+class WebRTCFFmpegBridge extends ScryptedDeviceBase implements BufferConverter {
+    constructor(public plugin: WebRTCPlugin, nativeId: string) {
+        super(nativeId);
+
+        this.fromMimeType = ScryptedMimeTypes.RTCSignalingChannel;
+        this.toMimeType = ScryptedMimeTypes.FFmpegInput;
+    }
+
+    async convert(data: any, fromMimeType: string, toMimeType: string, options?: MediaObjectOptions): Promise<any> {
+        const channel = data as RTCSignalingChannel;
```
koush, line 796: This also looks copy pasted from other code, should consolidate.

plugins/webrtc/src/main.ts
```diff
@@ -727,15 +715,17 @@ export async function fork() {
     }
 }
 
-class WebRTCBridge extends ScryptedDeviceBase implements BufferConverter {
+class WebRTCBridge extends ScryptedDeviceBase implements MediaConverter {
```
koush, line 718: it looks like this was consolidated into webrtc bridge, which is itself just a buffer converter. the plugin is a buffer converter as well. let's consolidate all 3.

koush: I actually added media converter for webrtc plugin and rebroadcast plugin to better handle multiple conversions. just had to wait a reasonable amount of time for servers to update.

koush:
awesome glad to finally clean this up

PR #1574, opened 2024-09-12
server: python createMediaManager func

bjia56:
Enables connectScryptedClient in the browser to access the API on behalf of Python plugins

### DIFF
--- a/server/python/plugin_remote.py
+++ b/server/python/plugin_remote.py
@@ -1043,6 +1043,11 @@ async def plugin_async_main(
         peer, api, pluginId, hostInfo, loop
     )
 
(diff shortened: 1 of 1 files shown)


server/python/plugin_remote.py
```diff
@@ -1043,6 +1043,11 @@ async def plugin_async_main(
         peer, api, pluginId, hostInfo, loop
     )
 
+    async def createMediaManager():
+        return MediaManager(await peer.getParam("getMediaManager"))
```
koush, line 1047: I think it should likely return the object as is, without the wrapper class. the caller will wrap it.

PR #1587, opened 2024-09-23
server: implement python listen + listenDevice

bjia56:
Appears to work fine when listening to a dummy switch's BinarySensor

### DIFF
--- a/server/python/plugin_remote.py
+++ b/server/python/plugin_remote.py
@@ -4,6 +4,7 @@ import asyncio
 import base64
 import gc
 import hashlib
+import inspect
 import multiprocessing
 import multiprocessing.connection
 import os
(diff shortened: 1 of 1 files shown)


koush:
signature is incorrect in that stub code, here's types.py:


```py

    def listen(self, callback: EventListener) -> EventListenerRegister:
        pass

    def listenDevice(self, id: str, event: str | EventListenerOptions, callback: EventListener) -> EventListenerRegister:
        pass
```

koush:
> Do you prefer listen/listenDevice to be synchronous? It makes the api portion of listenDevice more complex, but doable.

yeah in the early days I did have listen/listenDevice/unregister asynchronous on the js side. for unregister it is weird in that events can come in after invocation but before return.
by making it sync, it guarantees the local callback is immediately disconnected, so any messages en route from server are dropped.
for python, it is clunky in that all invocations from listen/listenDevice in the constructor (which is sync) must have an asyncio.runcoroutine launch pattern.

PR #1595, opened 2024-10-02
core: clear local url converter cache after 10m

bjia56:
Stopgap fix for excessive memory usage when something repeatedly converts to a LocalUrl media type

### DIFF
--- a/plugins/core/src/converters.ts
+++ b/plugins/core/src/converters.ts
@@ -59,6 +59,7 @@ export class BufferHost extends ScryptedDeviceBase implements HttpRequestHandler
         const filename = uuid + (extension ? `.${extension}` : '');
 
         this.hosted.set(`/${filename}`, { data: buffer, fromMimeType, toMimeType });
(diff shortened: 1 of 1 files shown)


koush:
can you check the usage of converting to this type in tree to verify other plugins aren't using it for long term urls? I don't think anything is off hand.

bjia56:
Usage I see:
- Converting text to audio in notifier-mixin: https://github.com/koush/scrypted/blob/adf4e797c12df535933745f26743dd4960ad5731/plugins/notifier-mixin/src/main.ts#L11
- Converting images/audio in chromecast: https://github.com/koush/scrypted/blob/adf4e797c12df535933745f26743dd4960ad5731/plugins/chromecast/src/main.ts#L176-L187

Seems like these are then sent to external services, which presumably will load the content near-immediately?

bjia56:
I'd be ok with extending the timeout to 5m or so as well, as you mentioned in Discord

koush:
yeah 10m is good

PR #1606, opened 2024-10-09
Reolink presets

apocaliss92:
Add presets to reolink
Fix PTZ movement rounded to 0
ONVIF configs moved to a subgroup


plugins/reolink/src/main.ts
```diff
@@ -81,7 +81,7 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
             hide: true,
         },
         ptz: {
-            subgroup: 'Advanced',
+            subgroup: 'ONVIF',
```
koush, line 84: reolink isn't using ONVIF, this setting group shouldn't be renamed

plugins/reolink/src/main.ts
```diff
@@ -94,6 +94,35 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
                 this.updatePtzCaps();
             },
         },
+        presets: {
+            subgroup: 'ONVIF',
```
koush, line 98: not onvif

plugins/reolink/src/main.ts
```diff
@@ -193,6 +241,33 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
 
     async ptzCommand(command: PanTiltZoomCommand): Promise<void> {
         const client = this.getClient();
+
+        const { movement, speed } = command;
+        if (movement) {
+            const onvifClient = await this.getOnvifClient();
```
koush, line 247: this shouldn't be using onvif.

koush:
the reolink plugin should use the reolink APIs to send ptz commands. if you need to make fixes to the ONVIF ptz, thats in the onvif plugin.

apocaliss92:
I thought the same after looking at the API, working on it, thanks !

PR #1608, opened 2024-10-09
Reolink presets

apocaliss92:


plugins/reolink/src/reolink-api.ts
```diff
@@ -294,7 +294,7 @@ export class ReolinkCameraClient {
             op += 'Up';
 
         if (op) {
-            await this.ptzOp(op, Math.round(Math.abs(command?.pan || command?.tilt || 1) * 10));
```
apocaliss92, line 266: I used ceil instead of round here because the commands from the app where < 0.5 and they were rounded to 0 instead of 1

plugins/reolink/src/reolink-api.ts
```diff
@@ -201,40 +213,59 @@ export class ReolinkCameraClient {
...
-        }, createReadable([
+        }, this.createReadable([
             {
                 cmd: "PtzCtrl",
                 param: {
                     channel: this.channelId,
                     op,
                     speed,
                     timeout: 1,
+                    id: id ? Number(id) : undefined
```
koush, line 254: make the id parameter itself a number and do the validation higher up.

plugins/reolink/src/reolink-api.ts
```diff
@@ -201,40 +213,59 @@ export class ReolinkCameraClient {
...
                 }
             },
         ]));
 
+        if(op === 'ToPos') {
```
koush, line 259: I think a separate ptzPreset would make more sense, since this code doesn't share all that much with preset usage (ie, the subsequent sleep and stop)

apocaliss92: yes makes definitely sense, done

plugins/reolink/src/reolink-api.ts
apocaliss92: done

plugins/reolink/src/reolink-api.ts
```diff
@@ -201,40 +213,59 @@ export class ReolinkCameraClient {
...
                     channel: this.channelId,
                     op,
                     speed,
                     timeout: 1,
+                    id
                 }
             },
         ]));
 
+        if (op === 'ToPos') {
```
koush, line 259: I think you can remove this now

apocaliss92: removed

PR #1609, opened 2024-10-11
server: aggressively catch child process pipe errors

bjia56:
In the case where a plugin requests a non-standard Python version that must be installed via portable-python, the extra pipes are used via PassThroughs and errors are not properly caught. This results in a high probability of a server crash when the plugin is killed and restarted.

### DIFF
--- a/server/src/plugin/runtime/python-worker.ts
+++ b/server/src/plugin/runtime/python-worker.ts
@@ -111,6 +111,13 @@ export class PythonRuntimeWorker extends ChildProcessWorker {
 
             this.worker.stdout.pipe(this.stdout);
             this.worker.stderr.pipe(this.stderr);
+
+            // aggressively catch errors
+            const onErr = (e: Error) => this.emit('error', e);
+            this.worker.stdout.on('error', onErr);
+            this.worker.stderr.on('error', onErr);
+            this.worker.stdio[3].on('error', onErr);
+            this.worker.stdio[4].on('error', onErr);
         };
 
         let pluginPythonVersion: string = options.packageJson.scrypted.pythonVersion?.[os.platform()]?.[os.arch()] || options.packageJson.scrypted.pythonVersion?.default;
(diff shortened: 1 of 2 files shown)


koush:
I suspect this may also affect node child processes. it's probably manifested but a systemd/docker restart brings the server back up immediately. it would explain why typescript debugging occasionally crashes the server too. can we get a general function in ChildProcessWorker that hooks all stdio for errors and emits?

koush:
incidentally I think the node pipe call should deliver errors down the pipe.

server/src/plugin/runtime/python-worker.ts
```diff
@@ -111,6 +111,14 @@ export class PythonRuntimeWorker extends ChildProcessWorker {
 
             this.worker.stdout.pipe(this.stdout);
             this.worker.stderr.pipe(this.stderr);
```
koush, line 113: I think the pipe call automatically sends this.worker.stderr errors to this.stderr.

bjia56: I believe either 3 or 4, the socket that raised the error had this specific event handler configured: https://github.com/nodejs/node/blob/9f9069d3130dc336a86271ca4e3c36e9b87cce90/lib/internal/child_process.js#L456

bjia56:
> I suspect this may also affect node child processes. it's probably manifested but a systemd/docker restart brings the server back up immediately. it would explain why typescript debugging occasionally crashes the server too. can we get a general function in ChildProcessWorker that hooks all stdio for errors and emits?

I moved this to setupWorker which seems to be the shared function called by subclasses

server/src/plugin/runtime/python-worker.ts
koush: gotcha, that makes sense, as 1 and 2 (stdout/stderr) are handled via pipe.

PR #1611, opened 2024-10-15
Force token on getAbility if the get call fails

apocaliss92:
More recent Reolink devices fail on using some call using the GET endpoint + username/password comination as authentication. This PR extract the api to get the token in a separated token data, to not collide with the other methods, and is used by an alternative getAbility method as documented in the official Reolink API documentation

### DIFF
--- a/plugins/reolink/src/probe.ts
+++ b/plugins/reolink/src/probe.ts
@@ -46,6 +46,43 @@ async function getDeviceInfo(host: string, username: string, password: string): 
     return response.body?.[0]?.value?.DevInfo;
 }
 
+export async function getToken(host: string, username: string, password: string) {
+    const url = new URL(`http://${host}/api.cgi`);
+    const params = url.searchParams;
+    params.set('cmd', 'Login');
+
+    const response = await httpFetch({
+        url,
+        method: 'POST',
+        responseType: 'json',
+        rejectUnauthorized: false,
+        body: [
+            {
+                cmd: 'Login',
+                action: 0,
+                param: {
+                    User: {
+                        userName: username,
+                        password: password
+                    }
+                }
+            },
+        ],
+    });
+
+    const token = response.body?.[0]?.value?.Token?.name || response.body?.value?.Token?.name;
+    if (!token)
+        throw new Error('unable to login');
+    const { body } = response;
+    const leaseTimeSeconds: number = body?.[0]?.value?.Token.leaseTime || body?.value?.Token.leaseTime;
+    return {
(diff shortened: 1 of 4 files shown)


plugins/reolink/src/probe.ts
```diff
@@ -46,6 +46,43 @@ async function getDeviceInfo(host: string, username: string, password: string):
     return response.body?.[0]?.value?.DevInfo;
 }
 
+export async function getToken(host: string, username: string, password: string) {
```
apocaliss92, line 49: This block is extracted from the `getLoginParameters` method to allow separated reuse

plugins/reolink/src/reolink-api.ts
```diff
@@ -148,6 +169,37 @@ export class ReolinkCameraClient {
         };
     }
 
+    async getAbilityWithToken() {
```
koush, line 172: rename to getAbility. call will always use token.

apocaliss92: You mean replace with the already existing getAbility?

koush:
there's too much copy pasting happening here. you replicated bot the getAbility and the getToken code. this can be implemented cleaner.

plugins/reolink/src/reolink-api.ts
```diff
@@ -85,6 +91,21 @@ export class ReolinkCameraClient {
         this.tokenLease = Date.now() + 1000 * leaseTimeSeconds;
     }
 
+    async requestWithToken(options: HttpFetchOptions<Readable>, body?: Readable) {
+        if (!this.tokenData?.token || !this.tokenData.leaseTime || this.tokenData.leaseTime > Date.now()) {
```
koush, line 95: what's happening here? requestWithLogin already does token based requests

koush: seems you can add a parameter to requestWithLogin to force a token based request. copy pasting the entire method seems unnecessary

koush:
I think what you want is a client mode that only operates with tokens and this always fetches a token from login. Trying to change the behavior throughout is going to be very messy.

apocaliss92:
> I think what you want is a client mode that only operates with tokens and thus always fetches a token from login. Trying to change the behavior throughout is going to be very messy.

Yes I didn't want to change to much the current logic, I think I will just fork out this and trying out on my system, thanks for the comments!

koush:
yeah, I think adding a token mode to the client would be trivial and also non instrusive. existing API calls work work as is.

PR #1613, opened 2024-10-18
Reolink token only requests

apocaliss92:


plugins/reolink/src/probe.ts
```diff
@@ -46,20 +46,21 @@ async function getDeviceInfo(host: string, username: string, password: string):
...
-export async function getLoginParameters(host: string, username: string, password: string) {
-    try {
-        await getDeviceInfo(host, username, password);
-        return {
-            parameters: {
-                user: username,
-                password,
-            },
-            leaseTimeSeconds: Infinity,
+export async function getLoginParameters(host: string, username: string, password: string, forceToken?: boolean) {
```
apocaliss92, line 49: user/pass skipped if token is forced

plugins/reolink/src/reolink-api.ts
```diff
@@ -73,25 +73,30 @@ export class ReolinkCameraClient {
         return pt;
     }
 
-    async login() {
-        if (this.tokenLease > Date.now()) {
+    async login(forceToken?: boolean) {
+        if ((!forceToken ? true : this.tokenLease !== Infinity) && this.tokenLease > Date.now()) {
```
apocaliss92, line 77: Add check, if the token is forced, make sure it's not Infinity otherwise it would always pass if the previous login was not forced

plugins/reolink/src/reolink-api.ts
```diff
@@ -73,25 +73,30 @@ export class ReolinkCameraClient {
...
 
-    async requestWithLogin(options: HttpFetchOptions<Readable>, body?: Readable) {
-        await this.login();
+    async requestWithLogin(options: HttpFetchOptions<Readable>, body?: Readable, forceToken?: boolean) {
+        await this.login(forceToken);
         const url = options.url as URL;
         const params = url.searchParams;
-        for (const [k, v] of Object.entries(this.parameters)) {
-            params.set(k, v);
+        if (forceToken) {
```
apocaliss92, line 92: Do not add username/password if forced to token, some calls fail

koush:
put the forceToken as a property of the client itself. So then the client operates in token mode for all requests. Then create the client in places where a token is required.

apocaliss92:
The problem with this approach is that reolink allows a maximum amount of session, if I create the client only in the calls needed, it will block the API very quickly

apocaliss92:
Maybe it can work having it on the camera level, like clientV2

koush:
> The problem with this approach is that reolink allows a maximum amount of session, if I create the client only in the calls needed, it will block the API very quickly

I'm not sure what you mean? When does it block? The token should be automatically refreshed based on lease time.

apocaliss92:
> > The problem with this approach is that reolink allows a maximum amount of session, if I create the client only in the calls needed, it will block the API very quickly
> 
> I'm not sure what you mean? When does it block? The token should be automatically refreshed based on lease time.

If I have to create the token only for the calls need it, it will require a new token every call, it must be persistent

apocaliss92:
Actually the changes are not only client, also on the login itself which is in the probe file

koush:
> > > The problem with this approach is that reolink allows a maximum amount of session, if I create the client only in the calls needed, it will block the API very quickly
> > 
> > 
> > I'm not sure what you mean? When does it block? The token should be automatically refreshed based on lease time.
> 
> If I have to create the enforced client only for the calls need it, it will require a new token every call, it must be persistent

What do you mean? The token has a lifetime of validity. I think it lasts like 15 minutes or something, and then must be refreshed.

PR #1614, opened 2024-10-19
Reolink only-token client added

apocaliss92:


plugins/reolink/src/reolink-api.ts
```diff
@@ -49,7 +49,7 @@ export class ReolinkCameraClient {
     parameters: Record<string, string>;
     tokenLease: number;
 
-    constructor(public host: string, public username: string, public password: string, public channelId: number, public console: Console) {
+    constructor(public host: string, public username: string, public password: string, public channelId: number, public console: Console, public forceToken?: boolean) {
```
koush, line 52: make this public readonly, forceToken can not be changed

plugins/reolink/src/reolink-api.ts
```diff
@@ -74,13 +74,13 @@ export class ReolinkCameraClient {
     }
 
     async login() {
-        if (this.tokenLease > Date.now()) {
+        if ((!this.forceToken ? true : this.tokenLease !== Infinity) && this.tokenLease > Date.now()) {
```
koush, line 77: this change is not necessary if forceToken is readonly

plugins/reolink/src/reolink-api.ts
```diff
@@ -89,8 +89,12 @@ export class ReolinkCameraClient {
         await this.login();
         const url = options.url as URL;
         const params = url.searchParams;
-        for (const [k, v] of Object.entries(this.parameters)) {
-            params.set(k, v);
+        if(this.forceToken) {
```
koush, line 92: this change is not necessary.

apocaliss92: done, some calls fail if providing both token and user/pass, but if forceToken is set they will never be set

plugins/reolink/src/reolink-api.ts
apocaliss92: done

plugins/reolink/src/reolink-api.ts
apocaliss92: done

plugins/reolink/src/reolink-api.ts
koush: How would both token and username pass be set?

apocaliss92: in this case it's not possible, it was something from the previous branch

PR #1620, opened 2024-10-21
Reolink add nvr flag

apocaliss92:


plugins/reolink/src/reolink-api.ts
```diff
@@ -201,16 +201,53 @@ export class ReolinkCameraClient {
...
         if (error) {
             this.console.error('error during call to getDeviceInfo', error);
             throw new Error('error during call to getDeviceInfo');
         }
-        return response.body?.[0]?.value?.DevInfo;
+
+        const deviceInfo: DevInfo = await response.body?.[0]?.value?.DevInfo;
+
+        // Will need to check if it's valid for NVR and NVR_WIFI
+        if (['HOMEHUB'].includes(deviceInfo.exactType)) {
```
koush, line 217: invert this logic so if it doesn't include, it immediately returns deviceInfo. Remaining code will be only for reolink home hub logic.

plugins/reolink/src/reolink-api.ts
```diff
@@ -201,16 +201,53 @@ export class ReolinkCameraClient {
         const url = new URL(`http://${this.host}/api.cgi`);
         const params = url.searchParams;
         params.set('cmd', 'GetDevInfo');
-        const response = await this.requestWithLogin({
+        let response = await this.requestWithLogin({
```
koush, line 204: why are these "let" and reused?

Use different variables for different request types. mutability on the response makes understanding this code later difficult.

plugins/reolink/src/reolink-api.ts
apocaliss92: done!

PR #1621, opened 2024-10-21
Reolink battery params supported

apocaliss92:


plugins/reolink/src/main.ts
```diff
@@ -318,10 +324,28 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
...
+        if (this.batteryTimeout) {
+            clearInterval(this.batteryTimeout);
+        }
+
+        this.batteryTimeout = setInterval(async () => {
+            const api = this.getClientWithToken();
+
+            const { batteryPercent, sleep } = await api.getBatteryInfo();
+            this.batteryLevel = batteryPercent;
+            this.online = !sleep;
```
apocaliss92, line 345: This sleep parameter is a very useful thing to keep to build more stuff later on, not sure if it's the right place where to store it? It tells if the camera is in standby. An use case I use it for is to generate a snapshot as soon as it comes online to save battery even more

plugins/reolink/src/main.ts
```diff
@@ -222,7 +223,7 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
         let abilities;
         try {
             abilities = await api.getAbility();
-        } catch(e) {
```
koush, line 225: undo

apocaliss92: done in both lines

plugins/reolink/src/main.ts
```diff
@@ -293,6 +294,11 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
             && this.storageSettings.values.abilities?.value?.Ability?.supportAudioAlarm?.ver !== 0;
     }
 
+    hasBattery() {
+        const batteryConfigVer = this.storageSettings.values.abilities?.value?.Ability?.abilityChn?.[this.getRtspChannel()].battery?.ver ?? 0;
```
koush, line 298: need optional index.

        const batteryConfigVer = this.storageSettings.values.abilities?.value?.Ability?.abilityChn?.[this.getRtspChannel()]?.battery?.ver ?? 0;

plugins/reolink/src/main.ts
```diff
@@ -318,10 +324,28 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
...
             interfaces.push(ScryptedInterface.DeviceProvider);
+        if (this.hasBattery()) {
+            interfaces.push(ScryptedInterface.Battery, ScryptedInterface.Online);
+            await this.startBatteryCheckInterval();
+        }
 
         await this.provider.updateDevice(this.nativeId, name, interfaces, type);
     }
 
+    async startBatteryCheckInterval() {
```
koush, line 335: I dont think this function needs to be async?

plugins/reolink/src/main.ts
```diff
@@ -318,10 +324,28 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
         }
         if (this.hasSiren())
             interfaces.push(ScryptedInterface.DeviceProvider);
+        if (this.hasBattery()) {
+            interfaces.push(ScryptedInterface.Battery, ScryptedInterface.Online);
+            await this.startBatteryCheckInterval();
```
koush, line 329: await not necessary, see comment below.

plugins/reolink/src/main.ts
```diff
@@ -318,10 +324,28 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
...
 
+    async startBatteryCheckInterval() {
+        if (this.batteryTimeout) {
+            clearInterval(this.batteryTimeout);
+        }
+
+        this.batteryTimeout = setInterval(async () => {
+            const api = this.getClientWithToken();
+
+            const { batteryPercent, sleep } = await api.getBatteryInfo();
```
koush, line 343: need failure handling.

plugins/reolink/src/main.ts
apocaliss92: wops, done

plugins/reolink/src/main.ts
apocaliss92: removed

plugins/reolink/src/main.ts
apocaliss92: removed

plugins/reolink/src/main.ts
apocaliss92: added

koush:
why close? haven't had time to review yet

PR #1622, opened 2024-10-21
Reolink floodlight device

apocaliss92:
This should be coming after the battery support PR, since it needs to check the battery interface to ping the camera to get the current light status. I'm expecting also conflicts from it.
I can just close it if needed, polish and open it afterwards


plugins/reolink/src/main.ts
```diff
@@ -287,6 +308,14 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
             && this.storageSettings.values.abilities?.value?.Ability?.supportAudioAlarm?.ver !== 0;
     }
 
+    hasFloodlight() {
```
koush, line 311: need otptional indexing and less copy pasting on the abilitychn.

koush:
why closed?

apocaliss92:
I will reopen rebased on the other, there will be some conflicts

PR #1625, opened 2024-10-24
feat(sdk): add support for notification channel in NotifierOptions

USA-RedDragon:
This change spins off from conversation at https://github.com/scryptedapp/homeassistant/pull/1 and allows consumers of the `Notifier` interface to specify a channel for notifications to be sent to. Android platforms can use this to send notifications to a specific channel, allowing the user to have fine-grained control over the audio and priority of the notifications they receive.

I'm not certain my `automation-actions.ts` edit is enough to make this work, I didn't readily see where these were consumed in the codebase.

Potential TODOs:

- Consumers of Notifier that might want to set a channel:
  * `manage.scrypted.app` - Should the frontend test notification set a channel of `testing` or should it stay with the general channel? https://github.com/koush/manage.scrypted.app/blob/168317f6790e7012560e7d2bb755c6fd03c0a308/src/components/interfaces/Notifier.vue#L97
  * `scryptedapp/notifier-zone-filter` - https://github.com/scryptedapp/notifier-zone-filter/blob/64d1d8130912ccab73c4e360db5cce757ccea1d6/src/main.py#L457 and https://github.com/scryptedapp/notifier-zone-filter/blob/64d1d8130912ccab73c4e360db5cce757ccea1d6/src/main.py#L460
  * `scryptedapp/securitysystem` - https://github.com/scryptedapp/securitysystem/blob/cd82bbde8a2414e738a035248e4646c3102e3221/src/main.ts#L129
  * Also, I'm not seeing where the results of the object detection are obtained in relation to conditional notifications, so that will still want to set a channel
- Implementations of Notifier that might be able to consume the channel:
  * Pushover? Edit: looks like pushover doesn't yet support this (https://support.pushover.net/i290-support-notification-channels-specific-to-application) - https://github.com/scryptedapp/pushover/blob/c1d28ca0feb2535d061e0b502e57fb707de3c17e/src/main.ts#L78
  * Home Assistant - PR - https://github.com/scryptedapp/homeassistant/pull/1

### DIFF
--- a/plugins/core/src/automation-actions.ts
+++ b/plugins/core/src/automation-actions.ts
@@ -81,6 +81,10 @@ addAction(ScryptedInterface.Notifier, {
         type: 'interface',
         deviceFilter: `deviceInterface === '${ScryptedInterface.VideoCamera}' || deviceInterface === '${ScryptedInterface.Camera}'`,
     },
+    notificationChannel: {
+        title: 'Channel (Android only)',
+        type: 'string',
+    },
 }, async function invoke(device: ScryptedDevice & Notifier, storageSettings) {
     let { notificationMediaUrl } = storageSettings;
     if (notificationMediaUrl && !notificationMediaUrl?.includes('://')) {
@@ -100,5 +104,6 @@ addAction(ScryptedInterface.Notifier, {
 
     return device.sendNotification(storageSettings.notificationTitle as string, {
         body: storageSettings.notificationBody as string,
+        channel: storageSettings.notificationChannel as string,
     }, notificationMediaUrl);
 });
--- a/plugins/core/src/automation-actions.ts
+++ b/plugins/core/src/automation-actions.ts
@@ -81,10 +81,6 @@ addAction(ScryptedInterface.Notifier, {
         type: 'interface',
         deviceFilter: `deviceInterface === '${ScryptedInterface.VideoCamera}' || deviceInterface === '${ScryptedInterface.Camera}'`,
     },
-    notificationChannel: {
-        title: 'Channel (Android only)',
-        type: 'string',
-    },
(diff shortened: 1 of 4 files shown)


sdk/types/src/types.input.ts
```diff
@@ -225,6 +225,7 @@ export interface NotifierOptions {
   badge?: string;
   bodyWithSubtitle?: string;
   body?: string;
+  channel?: string;
```
koush, line 228: maybe:

```ts
android?: {
  channel?: string;
}
```

plugins/core/src/automation-actions.ts
```diff
@@ -81,6 +81,10 @@ addAction(ScryptedInterface.Notifier, {
         type: 'interface',
         deviceFilter: `deviceInterface === '${ScryptedInterface.VideoCamera}' || deviceInterface === '${ScryptedInterface.Camera}'`,
     },
+    notificationChannel: {
```
koush, line 84: I don't think this should be in the automation ui, since it will almost certainly never be used. It can remain accessible from an automation script via the API.

sdk/types/src/types.input.ts
USA-RedDragon: https://github.com/koush/scrypted/pull/1625/commits/9edb752d133eedf9749703a7ecd437b81f6f8f89

plugins/core/src/automation-actions.ts
USA-RedDragon: https://github.com/koush/scrypted/pull/1625/commits/564283e14f78492b97f5986a5cbffb63f63c6323

koush:
looks good

koush:
published

PR #1626, opened 2024-10-25
Add flag to not autoenable devices on creation

apocaliss92:
Same changes of https://github.com/koush/scrypted/pull/1615 for Homekit


plugins/homekit/src/main.ts
```diff
@@ -111,6 +111,12 @@ export class HomeKitPlugin extends ScryptedDeviceBase implements MixinProvider,
             hide: true,
             description: 'The last home hub to request a recording. Internally used to determine if a streaming request is coming from remote wifi.',
         },
+        disableAutoAdd: {
```
koush, line 114: dont use negative names as variables. should be called `autoAdd`

apocaliss92: done

PR #1632, opened 2024-10-29
add raspberry pi 5 model no

greg6775:


koush:
there's no hardware decode/encode on pi 5.

PR #1633, opened 2024-10-30
Reolink floodlighs

apocaliss92:
Add support to reolink lights


plugins/reolink/src/main.ts
```diff
@@ -295,6 +316,14 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
             && this.storageSettings.values.abilities?.value?.Ability?.supportAudioAlarm?.ver !== 0;
     }
 
+    hasFloodlight() {
+        const channel = this.getRtspChannel();
+        const floodLightConfigVer = this.storageSettings.values.abilities?.value?.Ability?.abilityChn?.[channel].floodLight?.ver ?? 0;
```
koush, line 321: optional chain the index. use the same variable so you're not evaluating the same thing 3 times.

plugins/reolink/src/main.ts
```diff
@@ -766,40 +795,71 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
...
+            return this.floodlight;
         }
     }
 
     async releaseDevice(id: string, nativeId: string) {
         if (nativeId.endsWith('-siren')) {
             delete this.siren;
-        }
+        } else
+            if (nativeId.endsWith('--floodlight')) {
```
koush, line 860: why 2 dashes

plugins/reolink/src/reolink-api.ts
```diff
@@ -431,6 +431,59 @@ export class ReolinkCameraClient {
...
+            responseType: 'json',
+        }, this.createReadable(body));
+
+        const error = response.body?.[0]?.error;
+        if (error) {
+            this.console.error('error during call to setWhiteLedState', JSON.stringify(body), error);
+        }
+    }
+
+    async getWhiteLedState() {
```
koush, line 464: this does not look to be called, and I dont see anywhere that this value is polled to update the state from outside of scrypted?

plugins/reolink/src/main.ts
apocaliss92: done

plugins/reolink/src/main.ts
apocaliss92: wops, fixed

plugins/reolink/src/reolink-api.ts
apocaliss92: yes was a leftover, removed

plugins/reolink/src/main.ts
```diff
@@ -295,6 +316,21 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
             && this.storageSettings.values.abilities?.value?.Ability?.supportAudioAlarm?.ver !== 0;
     }
 
+    hasFloodlight() {
+        const channel = this.getRtspChannel();
+
+        const channelData = this.storageSettings.values.abilities?.value?.Ability?.abilityChn?.[channel];
+        if (channelData) {
+            const floodLightConfigVer = channelData.floodLight?.ver ?? 0;
```
koush, line 324: you need optional chaining on channelData.

apocaliss92: It's checked in the if, isn't it?

PR #1635, opened 2024-11-02
Add zoom RTSP streams to trackmix cameras

apocaliss92:


plugins/reolink/src/main.ts
```diff
@@ -688,6 +688,20 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
                 name: '',
                 id: 'autotrack.bcs',
```
koush, line 689: is this autotrack.bcs a valid stream? or just main.bcs and sub.bcs? verify resolution on all 3.

JoshuaSeidel: I know main and sub is, and those resolutions are right. Not sure about autocracy.bcs. 

How ever hitting those rtmps are still not working through scrypted

PR #1638, opened 2024-11-05
tapo: provide settings to override tapo camera's ip

latel:
# Motivation
I purchased a Tapo C120 camera, but after connecting it to Scrypted and trying various parameter combinations, the video still stutters. However, if I first connect it to Go2RTC and then integrate it into Scrypted using RTSP, it streams smoothly, though this approach sacrifices two-way audio functionality.

# Prompts
Since @scrypted/tapo’s two-way audio implementation actually relies on a less strict vulnerability in the Tapo camera, I suggest adding an option in the @scrypted/tapo plugin’s settings to specify the camera’s IP address. This way, even if a camera is connected through Go2RTC, the @scrypted/tapo plugin could still ensure connection to the actual Tapo camera IP. Given that the Go2RTC server IP and the Tapo camera IP are obviously different, in this scenario, @scrypted/tapo should not default to using Go2RTC’s virtual camera IP.

### DIFF
--- a/plugins/tapo/src/main.ts
+++ b/plugins/tapo/src/main.ts
@@ -11,6 +11,12 @@ class TapoIntercomMixin extends SettingsMixinDeviceBase<VideoCamera & Settings> 
             title: 'Cloud Password',
             description: 'The Tapo Cloud account password. This is not the same as the ONVIF/RTSP local password.',
             type: 'password',
+        },
+        cameraIP: {
+            title: 'Camera IP Address',
+            description: 'The actual IP address of your Tapo camera, Defaults to the one Scrypted provides.',
+            type: 'string',
+            placeholder: 'e.g., 192.168.1.100',
         }
     });
     client: Promise<TapoAPI>;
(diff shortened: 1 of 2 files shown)


plugins/tapo/src/main.ts
```diff
@@ -11,6 +11,12 @@ class TapoIntercomMixin extends SettingsMixinDeviceBase<VideoCamera & Settings>
             title: 'Cloud Password',
             description: 'The Tapo Cloud account password. This is not the same as the ONVIF/RTSP local password.',
             type: 'password',
+        },
+        cameraIP: {
+            title: 'Camera IP Address',
+            description: 'The actual IP address of your Tapo camera, Defaults to the one Scrypted provides.',
+            type: 'string',
+            placeholder: 'e.g., 192.168.1.100',
```
koush, line 19: remove the e.g.,

koush:
I don't think this is the correct fix. The correct fix is to determine why the feed is stuttering. It's likely due to something with the stream timestamps. Use the ffmpeg plugin to determine the ffmpeg arguments that provide a clean ingestion of the stream, possibly sanitizing timestamps.

latel:
image

There will be a new config entry added to specify the camera's real IP

ricmik:
Will this work for Taco cameras that are blocked from accessing the internet too?

latel:
> Will this work for Taco cameras that are blocked from accessing the internet too?

It works

PR #1650, opened 2024-11-26
Remove RTMP streams for nvr connected cams

apocaliss92:


plugins/reolink/src/main.ts
```diff
@@ -703,21 +703,30 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
         // anecdotally, encoders of type h265 do not have a working RTMP main stream.
         const mainEncType = this.storageSettings.values.abilities?.value?.Ability?.abilityChn?.[rtspChannel].mainEncType?.ver;
 
-        if (live === 2) {
-            if (mainEncType === 1) {
-                streams.push(rtmpSub, rtspMain, rtspSub);
+        const isNvr = this.client.getIsNvr(deviceInfo);
+
+        if (isNvr) {
```
apocaliss92, line 708: Interesting comparing, this is the only block added here (708-719), the non nvr branch is not changed at all

koush:
is there a better way to do this, to query if rtsp is available somehow.

apocaliss92:
Actually the streams are available but are broken when thorough NVR, we guided 2 users already this week to switch from rtmp to rtsp, maybe would just be better to set the defaults to rtsp? In that case what would be the criteria to let Scrypted pick them instead of rtmp, just order in the streams array?

koush:
> Actually the streams are available but are broken when thorough NVR, we guided 2 users already this week to switch from rtmp to rtsp, maybe would just be better to set the defaults to rtsp? In that case what would be the criteria to let Scrypted pick them instead of rtmp, just order in the streams array?

The issue with RTSP is that it is buggy on many reolink cameras causing dropouts or camera crashes. Furthermore RTSP only offers 2 streams and not 3 like RTMP.

PR #1654, opened 2024-12-05
Update Windows install script workaround npm issue

longzheng:
I was just testing the Windows install script recently and noticed that it would actually fail on the `npx` commands

After some digging it's due to Node.js removing creating the `%APPDATA%\npm` folder in the MSI installer https://github.com/nodejs/node/commit/0ae8bf8dbc3d5db9a604c7be66aa22893e109fa0 which then causes the `npx` command to fail https://github.com/nodejs/node/issues/53538

This was fixed in `npm` https://github.com/npm/cli/pull/7640 but it's not included with the npm installed in Node 20.11.1 specified in the install script

This `npm i -g npm` is a workaround https://github.com/nodejs/node/issues/53538#issuecomment-2223853437

Also added `#Requires -RunAsAdministrator` so the script will error if it's not run under admin

### DIFF
--- a/install/local/install-scrypted-dependencies-win.ps1
+++ b/install/local/install-scrypted-dependencies-win.ps1
@@ -26,6 +26,8 @@ $SCRYPTED_WINDOWS_PYTHON_VERSION="-3.9"
 # Refresh environment variables for py and npx to work
 $env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User") 
 
+node -v
+
 # Workaround Windows Node no longer creating %APPDATA%\npm which causes npx to fail
 # Fixed in newer versions of NPM but not the one bundled with 20.11.1
 # https://github.com/nodejs/node/issues/53538
@@ -102,7 +104,7 @@ svc.on("install", () => {
    setTimeout(() => {
      console.log("Starting service");
      svc.start();
-   }, 5000);
+   }, 10000);
 });
 svc.on("start", () => {
   console.log("Service started");
--- a/install/local/install-scrypted-dependencies-win.ps1
+++ b/install/local/install-scrypted-dependencies-win.ps1
@@ -1,3 +1,5 @@
+#Requires -RunAsAdministrator
+
 # Set-PSDebug -Trace 1
 
 # stop existing service if any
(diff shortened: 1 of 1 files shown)


koush:
@longzheng just to confirm, this still installs/runs the service itself as user account right?

longzheng:
@koush Correct, this should keep the service as user account.

However I'm a bit confused why
- the test failed https://github.com/koush/scrypted/actions/runs/12178053669/job/33967189873?pr=1654
- why previous tests without this change works (my expectation is that it shouldn't) e.g. https://github.com/koush/scrypted/actions/runs/12173418457/job/33953725454

longzheng:
Hmm I confirmed in a fresh VM that the new script fixes the `npx` error

image

And the server still works

image

Now I'm extra confused

longzheng:
That is so weird, removing the `npm i -g npm` fixes the problem for the test, but I can't replicate that in my VM.

koush:
Maybe the easier fix is to simply create the missing directory?

longzheng:
Yeah good idea. The fact that I can't repro it is kind of weird, I wonder how else is the test runner different.

longzheng:
Although if the server still runs even with that command failing, is the command necessary? How can the service/server run without the install?

koush:
The npx command that runs it will also install it if it's missing. Used to handle updates too.

longzheng:
> The npx command that runs it will also install it if it's missing. Used to handle updates too.

Do you think then maybe it's simpler to _not_ run the `npx` command in the install script and just let the `serve` service handle it?

koush:
the reason the install is done is because if would cause a long first run after install. and installation failures would be opaque.

longzheng:
OK that's fair enough.

I'm super curious then why the test server would fail with this, I need to get the logs. I might try modify the workflow to dump the output of the daemon logs.

longzheng:
@koush sorry to be a bother could you kill this run since it'll go on forever https://github.com/koush/scrypted/actions/runs/12193672987/job/34016265645

longzheng:
I logged `npm -v` before the script installs Node and I figured out the reason why the test server doesn't have the NPM problem, it's because it has NPM built-in to the image (and it's version `v18.20.5` so it doesn't have the installer change), so the `%APPDATA%\npm` folder must exist by default

https://github.com/koush/scrypted/actions/runs/12194017927/job/34017187505?pr=1654#step:4:12

longzheng:
OK I finally figured out what's happening.

Before the change:
- test server has Node 18
- install script did not upgrade to Node 20
- the server runs correctly

After the `npm i -g npm` change:
- test server still has Node 18
- npm was updated to latest version
- latest version of npm includes a [`npx.ps1` file for Powershell](https://github.com/npm/cli/blob/latest/bin/npx.ps1) (whereas the `npm` installed on the server only had `npm.cmd`)
- the `$NPX_PATH = (Get-Command npx).Path` command would return `npx.ps1` since we're in a Powershell install script
- The `service.js` was trying to spawn a child process with `npx.ps1` which does not work since the shell is not Powershell

New change:
- replace the `$NPX_PATH_ESCAPED` `.ps1` with `.cmd`

PR #1668, opened 2024-12-28
server: include CPython's `telnetlib` code directly

bjia56:
Python 3.13 removes `telnetlib` from the standard distribution as part of [PEP 594](https://peps.python.org/pep-0594/). We currently use the module in the Python REPL.

As [described in the Python forums](https://discuss.python.org/t/pep-594-take-2-removing-dead-batteries-from-the-standard-library/13508/59), it should be acceptable to copy it into other projects given the permissive license (which I have added to the source file).

Also bumps portable-python to the latest version with Python 3.13 availability.

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@scrypted/server",
-  "version": "0.123.68",
+  "version": "0.123.69",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@scrypted/server",
-      "version": "0.123.68",
+      "version": "0.123.69",
       "hasInstallScript": true,
       "license": "ISC",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


koush:
Can't this be done as part of the portable python install?

koush:
plugin repl can try import telnetlib, and then fallback somehow to using an included portable telnetlib in pypi or similar which must be specified in the requirements.txt? I'd consider ways to handle it that don't copy this file wholesale. my quick read is that telnetlib3 is also a drop in replacement.

bjia56:
I'd rather not bundle the old telnetlib with newer portable python versions, it wouldn't be a long term solution since I assume scrypted will eventually move to 3.13+ system python when ubuntu bumps.

It seems like the better approach here would be to switch to telnetlib3 from pypi and add it as a server dependency. It won't be a drop in replacement since telnetlib3 is asyncio and stdlib telnetlib is synchronous, however I can port our code over.

koush:
hmm ok let me think about this some more, I feel like there's gotta be a better solution here. I am also installing some stuff via pip already like debugpy:

https://github.com/koush/scrypted/blob/0cf48023853f08744b1e642272107171589701c2/server/bin/packaged-python.js#L20

Maybe something similar could be an option?

bjia56:
Yeah adding it to plugin_remote.py sounds like a good approach

bjia56:
This seems promising: https://pypi.org/project/standard-telnetlib/

PR #1674, opened 2025-01-04
Removed unreachable block

mike-marcacci:
While familiarizing myself with the architecture of this project I noticed that this block is unreachable and handled above. Figured I'd submit a quick fix.

Awesome project BTW.


koush:
This is reachable on RPI 4. RPI 5 does not have hardware acceleration.

koush:
it's been a minute since I've run this on an RPI4 , so let me know if I'm incorrect. But the mmal code path definitely existed on rpi4 at the time of writing.

koush:
nm only glanced at code, @bjia56 cleared it up

PR #1698, opened 2025-01-19
hikvision: add ERI-K104-P4 to the list of NVRs that doesn't support channel cap checks

gtalusan:
I have an ERI-K104-P4 NVR - which is only a couple of years old, but doesn't support these channel capability checks.  This lets Scrypted fallback to H264/AAC similar to the DS-76098NI-E2.

Works great now with a 4 channel set up.


gtalusan:
@koush thanks for the quick merge.  Do you mind publishing an artifact with this change?  I have a couple of different installations that need this fix.  Thanks, and thanks again for the quick responses on Discord.

PR #1699, opened 2025-01-20
homekit: Added Common Humidifer/Dehumidifer Settings to Thermostat and Fans

nberardi:
I am using Scrypted to pull my thermostat settings into Home Assistant via the HomeKit Controller and noticed that they require the following characteristics to be set to show the `Fan Mode` and `Humidity` settings on their climate card.  

* `Characteristic.TargetFanState` as [seen here](https://github.com/home-assistant/core/blob/dev/homeassistant/components/homekit_controller/climate.py#L161) to enable Fan Modes.  
* `Characteristic.CurrentFanState` as [seen here](https://github.com/home-assistant/core/blob/dev/homeassistant/components/homekit_controller/climate.py#L674) to indicate if the fan is active
* `Characteristic.TargetRelativeHumidity` as [seen here](https://github.com/home-assistant/core/blob/dev/homeassistant/components/homekit_controller/climate.py#L633) to enable Humidity card.
* `Characteristic.CurrentRelativeHumidity` as [seen here](https://github.com/home-assistant/core/blob/dev/homeassistant/components/homekit_controller/climate.py#L626) to indicate humidity to card.

I also took the opportunity to move the humidity settings to a common setup method for both `Thermostat` and `Fan`/

### DIFF
--- a/plugins/homekit/src/types/common.ts
+++ b/plugins/homekit/src/types/common.ts
@@ -1,4 +1,4 @@
-import sdk, { AirQuality, AirQualitySensor, CO2Sensor, DeviceProvider, Fan, FanMode, NOXSensor, OnOff, PM10Sensor, PM25Sensor, ScryptedDevice, ScryptedDeviceType, ScryptedInterface, VOCSensor } from "@scrypted/sdk";
+import sdk, { AirQuality, AirQualitySensor, CO2Sensor, DeviceProvider, Fan, FanMode, HumidityMode, HumiditySensor, HumiditySetting, HumiditySettingStatus, NOXSensor, OnOff, PM10Sensor, PM25Sensor, ScryptedDevice, ScryptedDeviceType, ScryptedInterface, VOCSensor } from "@scrypted/sdk";
 import { bindCharacteristic } from "../common";
 import { Accessory, Characteristic, CharacteristicEventTypes, Service, uuid } from '../hap';
 import type { HomeKitPlugin } from "../main";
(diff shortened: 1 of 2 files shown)


plugins/homekit/src/types/common.ts
```diff
@@ -96,24 +96,161 @@ export function addCarbonDioxideSensor(device: ScryptedDevice & CO2Sensor, acces
...
+        callback();
+        device.setHumidity({
+            humidifierSetpoint: value as number,
+        });
+    });
+
+    return service;
+}
+
+export function addDehumidifer(device: ScryptedDevice & HumiditySetting & HumiditySensor, accessory: Accessory): Service {
```
koush, line 207: I don't think this needs export

plugins/homekit/src/types/common.ts
```diff
@@ -96,24 +96,161 @@ export function addCarbonDioxideSensor(device: ScryptedDevice & CO2Sensor, acces
...
+        return 0;
+    }
+
+    bindCharacteristic(device, ScryptedInterface.HumiditySetting, service, Characteristic.TargetRelativeHumidity,
+        () => targetHumidity(device.humiditySetting));
+
+    return service;
+}
+
+export function addHumidifier(device: ScryptedDevice & HumiditySetting & HumiditySensor, accessory: Accessory): Service {
```
koush, line 192: I don't think this needs export

plugins/homekit/src/types/common.ts
nberardi: Circling back to this PR, since I missed the original comments... would you like me to submit a new PR for removing the `export`?

plugins/homekit/src/types/common.ts
nberardi: Circling back to this PR, since I missed the original comments... would you like me to submit a new PR for removing the `export`?

PR #1717, opened 2025-01-31
Improve snapshot handling for battery cams

apocaliss92:
Snapshots on Reolink battery cameras drains the battery. Only current solution is to use a dummy url in the snapshot plugin
These changes will limit the snapshots taken from the camera, avoiding to waking it up eccessively
The snap will be allowed only when the camera wakes up for other purposes, or every n seconds defined in the new setting added


koush:
this kind of stuff should be done in snapshot plugin

apocaliss92:
Let me try to dissect it then

koush:
Theres similar code in rebroadcast for handling battery cams.

PR #1718, opened 2025-02-01
Preserve battery on snapshots

apocaliss92:
Rebroadcast plugin: don't implement Online interface if the extended device implements it already
Snapshot: if battery camera, don't execute any snapshot if the camera is not online. Regularly check for online status of the camera  to take a snapshot as soon as the camera gets online again. Never allow prebuffer

### DIFF
--- a/plugins/prebuffer-mixin/src/main.ts
+++ b/plugins/prebuffer-mixin/src/main.ts
@@ -1754,7 +1754,9 @@ export class RebroadcastPlugin extends AutoenableMixinProvider implements MixinP
       return;
     if (!interfaces.includes(ScryptedInterface.VideoCamera))
       return;
-    const ret = [ScryptedInterface.VideoCamera, ScryptedInterface.Settings, ScryptedInterface.Online, REBROADCAST_MIXIN_INTERFACE_TOKEN];
+    const ret = [ScryptedInterface.VideoCamera, ScryptedInterface.Settings, REBROADCAST_MIXIN_INTERFACE_TOKEN];
+    if (!interfaces.includes(ScryptedInterface.Online))
+      ret.push(ScryptedInterface.Online)
     return ret;
   }
 
--- a/plugins/snapshot/src/main.ts
+++ b/plugins/snapshot/src/main.ts
@@ -2,7 +2,7 @@ import { AutoenableMixinProvider } from "@scrypted/common/src/autoenable-mixin-p
 import { AuthFetchCredentialState, authHttpFetch } from '@scrypted/common/src/http-auth-fetch';
 import { RefreshPromise, TimeoutError, createMapPromiseDebouncer, singletonPromise, timeoutPromise } from "@scrypted/common/src/promise-utils";
 import { SettingsMixinDeviceBase, SettingsMixinDeviceOptions } from "@scrypted/common/src/settings-mixin";
-import sdk, { BufferConverter, Camera, DeviceManifest, DeviceProvider, FFmpegInput, HttpRequest, HttpRequestHandler, HttpResponse, MediaObject, MediaObjectOptions, MixinProvider, RequestMediaStreamOptions, RequestPictureOptions, ResponsePictureOptions, ScryptedDevice, ScryptedDeviceType, ScryptedInterface, ScryptedMimeTypes, Setting, SettingValue, Settings, VideoCamera, WritableDeviceState } from "@scrypted/sdk";
+import sdk, { Battery, BufferConverter, Camera, DeviceManifest, DeviceProvider, FFmpegInput, HttpRequest, HttpRequestHandler, HttpResponse, MediaObject, MediaObjectOptions, MixinProvider, Online, RequestMediaStreamOptions, RequestPictureOptions, ResponsePictureOptions, ScryptedDevice, ScryptedDeviceType, ScryptedInterface, ScryptedMimeTypes, Setting, SettingValue, Settings, VideoCamera, WritableDeviceState } from "@scrypted/sdk";
 import { StorageSettings } from "@scrypted/sdk/storage-settings";
 import https from 'https';
 import os from 'os';
@@ -117,9 +117,23 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
     static lastGeneratedErrorImageTime = 0;
     lastAvailablePicture: Buffer;
     psos: ResponsePictureOptions[];
+    isBattery: boolean;
+    currentPictureOptions: RequestPictureOptions;
+    batteryCheckInterval: NodeJS.Timeout;
 
     constructor(public plugin: SnapshotPlugin, options: SettingsMixinDeviceOptions<Camera>) {
         super(options);
+
+        this.isBattery = this.mixinDeviceInterfaces.includes(ScryptedInterface.Battery);
+
+        if (this.isBattery) {
+            this.storageSettings.settings.snapshotsFromPrebuffer.value = 'Disabled';
+            this.storageSettings.settings.snapshotsFromPrebuffer.hide = true;
+
+            this.batteryCheckInterval = setInterval(async () => {
+                this.takePictureRaw(this.currentPictureOptions, true);
+            }, 15 * 1000);
+        }
     }
 
     get debugConsole() {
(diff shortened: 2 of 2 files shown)


plugins/prebuffer-mixin/src/main.ts
```diff
@@ -1754,7 +1754,9 @@ export class RebroadcastPlugin extends AutoenableMixinProvider implements MixinP
       return;
     if (!interfaces.includes(ScryptedInterface.VideoCamera))
       return;
-    const ret = [ScryptedInterface.VideoCamera, ScryptedInterface.Settings, ScryptedInterface.Online, REBROADCAST_MIXIN_INTERFACE_TOKEN];
+    const ret = [ScryptedInterface.VideoCamera, ScryptedInterface.Settings, REBROADCAST_MIXIN_INTERFACE_TOKEN];
+    if (!interfaces.includes(ScryptedInterface.Online))
```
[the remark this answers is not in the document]
apocaliss92: Maybe I will check both interfaces? If camera has both online and battery, the online won't be added, that would be the safest in terms of breaking changes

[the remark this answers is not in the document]
apocaliss92: done

plugins/snapshot/src/main.ts
```diff
@@ -117,35 +117,60 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
...
-                break;
-            case 'Disabled':
-                usePrebufferSnapshots = false;
-                break;
-            default:
-                // default behavior is to use a prebuffer snapshot if there's no camera interface and
-                // no explicit snapshot url.
-                if (!this.mixinDeviceInterfaces.includes(ScryptedInterface.Camera) && !this.storageSettings.values.snapshotUrl)
+        if (this.isBattery) {
+            usePrebufferSnapshots = false;
```
koush, line 157: This ends up completely ignoring the user selection.

plugins/snapshot/src/main.ts
```diff
@@ -117,35 +117,60 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
...
+                case 'true':
+                case 'Enabled':
                     usePrebufferSnapshots = true;
-                break;
+                    break;
+                case 'Disabled':
+                    usePrebufferSnapshots = false;
+                    break;
+                default:
+                    // default behavior is to use a prebuffer snapshot if there's no camera interface and
```
koush, line 168: do the battery check here.

plugins/snapshot/src/main.ts
apocaliss92: I have completely removed the setting to prebuffer, prebuffer in any shape on battery cameras should not be active
Should we still allow user to keep it on?

plugins/snapshot/src/main.ts
```diff
@@ -117,35 +117,60 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
...
+    isBattery: boolean;
+    currentPictureOptions: RequestPictureOptions;
+    batteryCheckInterval: NodeJS.Timeout;
 
     constructor(public plugin: SnapshotPlugin, options: SettingsMixinDeviceOptions<Camera>) {
         super(options);
+
+        this.isBattery = this.mixinDeviceInterfaces.includes(ScryptedInterface.Battery);
+
+        if (this.isBattery) {
```
apocaliss92, line 129: Here I have disabled the prebuffer by default, what do you think

apocaliss92:
Honestly I would disable the possibility to prebuffer even on the rebroadcast plugin..

plugins/snapshot/src/main.ts
```diff
@@ -117,35 +117,60 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
...
+            this.storageSettings.settings.snapshotsFromPrebuffer.value = 'Disabled';
+            this.storageSettings.settings.snapshotsFromPrebuffer.hide = true;
+
+            this.batteryCheckInterval = setInterval(async () => {
+                if (Date.now() - this.currentPictureTime < 60 * 1000)
```
koush, line 134: this is odd, shouldn't use polling for picture. won't this fail most of the time?

plugins/snapshot/src/main.ts
```diff
@@ -272,7 +297,8 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
         throw new Error('Snapshot Unavailable (Snapshot URL empty)');
     }
 
-    async takePictureRaw(options?: RequestPictureOptions): Promise<Buffer> {
+    async takePictureRaw(options?: RequestPictureOptions, ignoreDebounce?: boolean): Promise<Buffer> {
+        this.currentPictureOptions = options;
```
koush, line 301: this should not be cached, it changes based on request.

plugins/snapshot/src/main.ts
apocaliss92: Should I move the snapshot refresh on the camera plugin?

plugins/snapshot/src/main.ts
apocaliss92: Restored

plugins/snapshot/src/main.ts
apocaliss92: done

plugins/snapshot/src/main.ts
apocaliss92: removed

plugins/snapshot/src/main.ts
apocaliss92: removed

plugins/snapshot/src/main.ts
```diff
@@ -117,9 +117,12 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
     static lastGeneratedErrorImageTime = 0;
     lastAvailablePicture: Buffer;
     psos: ResponsePictureOptions[];
+    isBattery: boolean;
```
koush, line 120: this doesn't need to be a property. just check when you need to

plugins/snapshot/src/main.ts
```diff
@@ -142,10 +145,14 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
...
             default:
                 // default behavior is to use a prebuffer snapshot if there's no camera interface and
-                // no explicit snapshot url.
-                if (!this.mixinDeviceInterfaces.includes(ScryptedInterface.Camera) && !this.storageSettings.values.snapshotUrl)
-                    usePrebufferSnapshots = true;
-                break;
+                // no explicit snapshot url. If battery disable
+                if (this.isBattery) {
+                    usePrebufferSnapshots = false;
+                } else {
```
koush, line 151: else if

plugins/snapshot/src/main.ts
```diff
@@ -142,10 +145,14 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
...
+                    if (!this.mixinDeviceInterfaces.includes(ScryptedInterface.Camera) && !this.storageSettings.values.snapshotUrl)
+                        usePrebufferSnapshots = true;
+                    break;
```
koush, line 154: need break and end of default block.

apocaliss92: done

plugins/snapshot/src/main.ts
apocaliss92: done

plugins/snapshot/src/main.ts
apocaliss92: removed

plugins/snapshot/src/main.ts
```diff
@@ -288,6 +288,15 @@ class SnapshotMixin extends SettingsMixinDeviceBase<Camera> implements Camera {
             type: 'source',
             event: options?.reason === 'event',
         }, eventSnapshot ? 0 : 4000, async () => {
+            // If battery cam and not online, skip otherwise will wake up'
+            const realDevice = systemManager.getDeviceById<Online>(this.id);
+            if (this.mixinDeviceInterfaces.includes(ScryptedInterface.Battery) && !realDevice.online) {
+                return {
```
koush, line 294: I don't think this is valid return value. it should `throw new Error('not taking snapshot for battery camera')`

plugins/prebuffer-mixin/src/stream-settings.ts
```diff
@@ -117,6 +117,10 @@ export function createStreamSettings(device: MixinDeviceBase<VideoCamera>) {
         if (!msos)
             return;
 
+        if (device.mixinDeviceInterfaces.includes(ScryptedInterface.Sleep)) {
```
apocaliss92, line 120: Are you fine with this?

plugins/prebuffer-mixin/src/stream-settings.ts
```diff
@@ -1,5 +1,5 @@
 import { getH264DecoderArgs } from "@scrypted/common/src/ffmpeg-hardware-acceleration";
-import { MixinDeviceBase, ResponseMediaStreamOptions, VideoCamera } from "@scrypted/sdk";
```
koush, line 2: you can remove all the changes from rebroadcast completely. this looks leftover.

koush: nevermind

plugins/prebuffer-mixin/src/stream-settings.ts
koush: this seems wrong? Why is the stream selection being removed here?

apocaliss92: Just an idea to let you check, to not enable prebuffer by default for Sleep implementation, I think that's will be the case most of the times?

plugins/snapshot/src/main.ts
koush: > I have completely removed the setting to prebuffer, prebuffer in any shape on battery cameras should not be active
oh I see.  the issue is that there are battery cams that are ALSO powered. Ie, the battery is only used if power gets lost.

so I dont think this is correct.

apocaliss92: Yes reatored already

plugins/prebuffer-mixin/src/stream-settings.ts
koush: the available streams returned need to be constant, should never change based on sleep availability

apocaliss92: Oh wait I meant to change the default ones

apocaliss92: I'll remove this soon

apocaliss92: done, removed

PR #1719, opened 2025-02-01
Battery cams api fixes

apocaliss92:
Improves handling of online state on Reolink battery camses, increasing the checking on the online state and allow the snapshot plugin to get more relevant snapshots
Fix for battery camera motion reporting, it comes as an event instead of Ai result


plugins/reolink/src/main.ts
```diff
@@ -78,7 +78,7 @@ class ReolinkCameraFloodlight extends ScryptedDeviceBase implements OnOff, Brigh
     }
 }
 
-class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, Reboot, Intercom, ObjectDetector, PanTiltZoom {
+class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, Reboot, Intercom, ObjectDetector, PanTiltZoom, Online {
```
koush, line 81: I believe this Online interface needs to be reported in onDeviceDiscovered as well.

plugins/reolink/src/main.ts
```diff
@@ -380,12 +380,19 @@ class ReolinkCamera extends RtspSmartCamera implements Camera, DeviceProvider, R
...
+                }
+                if (batteryPercent !== this.batteryLevel) {
+                    this.batteryLevel = batteryPercent;
+                }
             }
             catch (e) {
                 this.console.log('Error in getting battery info', e);
             }
-        }, 1000 * 60 * 30);
+        }, 1000 * 10);
```
koush, line 395: this seems excessively frequent.

koush: doesn't making this request keep the battery camera awake?

plugins/reolink/src/main.ts
apocaliss92: I did it already long back, just forgot the interface here

plugins/reolink/src/main.ts
apocaliss92: No this won't wake it up, this will be handled by the hub and battery cams only work with hub on scrypted
What value would you suggest?

koush: I think preserving existing interval for non battery cams is preferable.

apocaliss92: The interval won't start for non battery cams at all

PR #1730, opened 2025-02-09
Reolink: Forward battery cams detections

apocaliss92:


koush:
this is mostly copy paste of another function?|

apocaliss92:
> this is mostly copy paste of another function?|

Yes to keep the other flow untouched, I can edit that one to fit both

koush:
I couldn't see much of a difference besides which endpoint was used, create a diff and can evaluate

apocaliss92:
Done

PR #1731, opened 2025-02-09
Sensors interface

apocaliss92:
I skipped all the changes to versions of package.json, I did it only in the py file
not sure if I missed anything

Used this as reference https://github.com/koush/scrypted/commit/5d87a1b2dd6b1d40a531a255df81985eed40be39

### DIFF
--- a/sdk/types/src/types.input.ts
+++ b/sdk/types/src/types.input.ts
@@ -218,8 +218,14 @@ export interface Buttons {
   buttons?: ('doorbell' | string)[];
 }
 
+export interface Sensor {
+  name: string;
+  value?: string | number;
+  unit?: string;
+}
+
 export interface Sensors {
-  sensors?: string[];
+  sensors: Record<string, Sensor>;
 }
 
 export interface PressButtons {
--- a/sdk/types/src/types.input.ts
+++ b/sdk/types/src/types.input.ts
@@ -218,6 +218,10 @@ export interface Buttons {
   buttons?: ('doorbell' | string)[];
 }
 
+export interface Sensors {
+  sensors?: string[];
+}
+
 export interface PressButtons {
   pressButton(button: string): Promise<void>;
 }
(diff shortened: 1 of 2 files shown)


apocaliss92:
Probably the single sensor could contain some metadata, i.e. unit or any descriptor of the utility, instead of plain strings

sdk/types/src/types.input.ts
```diff
@@ -218,6 +218,16 @@ export interface Buttons {
   buttons?: ('doorbell' | string)[];
 }
 
+export interface Sensor {
```
koush, line 221: name is required? what names are you using?

apocaliss92: For homeassistant, the id would be the entity_id, which could be pragmatic, friendly_name can give a bit of context. Most of the times entity IDs are autogenerated and not comprehensible

koush: gotcha, are there any other things you think might be necessary on the interface? 

my thinking is if it starts duplicating properties from all of ScryptedDevice it may just be better to use a sub device.

koush:
does your use case consolidate sensors onto a single device? like thermometer and humidity. Or is everything going to be separate devices with a single item in Sensors

apocaliss92:
The first use case on Scrypted would be to complete the tempest plugin and provide all those properties to this interface (something like 20 different values) into one device.
After this I was thinking about exposing a selector of homeassistant devices and put all the properties of those in a single device

sdk/types/src/types.input.ts
apocaliss92: I think this is enough, the unit can tell almost everything on how to treat the values
Scrypted need something like this imho to avoid have thousand of different devices olding just 1 value

apocaliss92:
Will be a task of consumers to treat these values, i.e. on the osd manager plugin I will let the user choose these devices and a subproperty of it

PR #1737, opened 2025-02-12
Add Supplemental Light & Alarm Support to Hikvision Plugin

akifbayram:
- Added supplemental light control (auto/manual mode, brightness control).
- Added alarm control (audio alarms, white light alarms, trigger settings).
Tested on Hikvision cameras with alarms/lights, with lights only, and with neither.


plugins/hikvision/src/main.ts
```diff
@@ -27,11 +28,319 @@ function channelToCameraNumber(channel: string) {
     return channel.substring(0, channel.length - 2);
 }
 
+export class HikvisionAlarmSwitch extends ScryptedDeviceBase implements OnOff, Settings {
```
koush, line 31: can you move these devices into separate files

koush:
Seems to be close to working, however my ir cameras do not work. the code assumes its always a colorvu cam.

PR #1744, opened 2025-02-22
Add polling to reolink discovered devices

apocaliss92:
- Adds polling to discovered devices, to check current state. Avoid the call if camera is sleeping, since all these commands are meant to actually wake up battery cams
- Remove on property initialization of devices to false, this would actually triggers everytime reset of the actual devices on plugin restart 

There could be a different call for older cameras on the Siren device, I could not test it out. Would go for the V2 now and check if any issue will be reported on older devices. There should be just a property version to check

### DIFF
--- a/plugins/reolink/src/main.ts
+++ b/plugins/reolink/src/main.ts
@@ -11,10 +11,21 @@ import { AIState, Enc, ReolinkCameraClient } from './reolink-api';
 
 class ReolinkCameraSiren extends ScryptedDeviceBase implements OnOff {
     sirenTimeout: NodeJS.Timeout;
+    sirenCheckInterval: NodeJS.Timeout;
 
     constructor(public camera: ReolinkCamera, nativeId: string) {
         super(nativeId);
-        this.on = false;
+
+        this.sirenCheckInterval = setInterval(async () => {
+            if (!this.camera.sleeping) {
+                const api = this.camera.getClient();
+                const { enabled } = await api.getSiren();
+
+                if (enabled !== this.on) {
+                    this.on = enabled;
+                }
+            }
+        }, 2000);
     }
 
     async turnOff() {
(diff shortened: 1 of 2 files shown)


plugins/reolink/src/main.ts
```diff
@@ -11,10 +11,21 @@ import { AIState, Enc, ReolinkCameraClient } from './reolink-api';
...
+
+        if (!this.camera.interfaces.includes(ScryptedInterface.Sleep)) {
+            this.sirenCheckInterval = setInterval(async () => {
+                const api = this.camera.getClient();
+                const { enabled } = await api.getSiren();
+
+                if (enabled !== this.on) {
+                    this.on = enabled;
+                }
+            }, 2000);
```
koush, line 27: I would use a async loop that runs forever with a 2 second sleep. setInterval can cause flooding if the request takes longer than 2 seconds. will crash camera.

apocaliss92: done, set it to 5 if it's fine, 10 is a very long time

koush:
It doesn’t seem like there’s any error handling if the request fail, which will cause poling to permanently stop. Would wrap every block in try catch

apocaliss92:
> It doesn’t seem like there’s any error handling if the request fail, which will cause poling to permanently stop. Would wrap every block in try catch

done!

gtalusan:
Hi @apocaliss92, this pull request seems to have broken the CX410 that I have configured with a siren.  The switch always appears to be ON.  Toggling it to OFF will automatically switch it back to ON - however the siren will never alarm.  Any idea what's going on?

apocaliss92:
@Is the camera behind hub?

gtalusan:
No, it isn't.  It's accessible directly by its own IP.

apocaliss92:
By any chance could you check if the homeassistant plugin shows the siren status correctly?

gtalusan:
I don't run HomeAssistant.

I only notice the state being sticky in HomeKit, and when I checked inside of Scrypted then the toggle switch for the siren was also sticky.

apocaliss92:
I think I got the issue, I will try to check it out tomorrow

gtalusan:
Great, thanks!

apocaliss92:
@gtalusan  any chance you could jump on discord and ping me? Or would be helpful to have the output of your camera as soon as you restart the reolink plugin, I'm mostly interested in the JSON data output right after the `getAbility` text

gtalusan:
@apocaliss92 here's a gist of getability from the camera in scrypted's logs: https://gist.github.com/gtalusan/a9fcbb26d02a4ea067a339fa581e2e95

I don't have Discord installed at the moment, but I can jump on today or tomorrow if you need more help.

PR #1745, opened 2025-02-23
add missing python3.12

lvnilesh:


koush:
plugins can get python version automatically via portable python.

lvnilesh:
This test fails without the proposed change

https://github.com/koush/scrypted/blob/977db49f873c30b4cf269a8610285f360c227e5d/install/docker/Dockerfile.full#L107C1-L108C1

koush:
likely need to update your BASE variable to noble.
the image built fine 2 weeks ago https://github.com/koush/scrypted/actions/runs/13228900705

PR #1764, opened 2025-03-05
Hikvision: Ptz + presets implemented

apocaliss92:
Not sure about all these formatting changes, I just ran the format hotkey on webstorm :| will revert if it's a problem

### DIFF
--- a/plugins/hikvision/src/main.ts
+++ b/plugins/hikvision/src/main.ts
@@ -558,18 +558,19 @@ export class HikvisionCamera extends RtspSmartCamera implements Camera, Intercom
                 this.console.error('Error in getPtzCapabilities', e);
             }
         } finally {
-            const hasPtz = !!ptzCapabilities.length;
-            this.hasPtz = hasPtz;
-            this.storage.setItem('hasPtz', hasPtz ? 'true' : 'false');
+            if (!!ptzCapabilities.length) {
+                this.hasPtz = true;
+                this.storage.setItem('hasPtz', 'true');
+            }
         }
     }
 
     updatePtzCaps(cameraPtzCapabilities: string[]) {
         this.ptzCapabilities = {
             ...this.ptzCapabilities,
-            pan: cameraPtzCapabilities?.includes('Pan'),
-            tilt: cameraPtzCapabilities?.includes('Tilt'),
-            zoom: cameraPtzCapabilities?.includes('Zoom'),
+            pan: this.ptzCapabilities?.pan || cameraPtzCapabilities?.includes('Pan'),
+            tilt: this.ptzCapabilities?.tilt || cameraPtzCapabilities?.includes('Tilt'),
+            zoom: this.ptzCapabilities?.zoom || cameraPtzCapabilities?.includes('Zoom'),
         }
     }
 
--- a/plugins/hikvision/src/main.ts
+++ b/plugins/hikvision/src/main.ts
@@ -593,7 +593,6 @@ export class HikvisionCamera extends RtspSmartCamera implements Camera, Intercom
 
         await this.updateDevice();
         this.updateDeviceInfo();
-        await this.fetchPresets();
     }
 
     async getOtherSettings(): Promise<Setting[]> {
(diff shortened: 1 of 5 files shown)


plugins/hikvision/src/main.ts
```diff
@@ -457,7 +482,7 @@ export class HikvisionCamera extends RtspSmartCamera implements Camera, Intercom
         return false;
     }
 
-    updateDevice() {
+    async updateDevice() {
```
koush, line 485: this should not be async, it should update the device with the current known state. figure out ptz state prior to calling this.

plugins/hikvision/src/main.ts
```diff
@@ -43,9 +43,15 @@ export class HikvisionCamera extends RtspSmartCamera implements Camera, Intercom
         super(nativeId, provider);
 
         this.hasSmartDetection = this.storage.getItem('hasSmartDetection') === 'true';
-        this.updateDevice();
-        this.updateDeviceInfo();
-        this.reportDevices();
+
+        (async () => {
```
koush, line 47: gut check says this change to async is problematic

plugins/hikvision/src/main.ts
```diff
@@ -478,12 +503,68 @@ export class HikvisionCamera extends RtspSmartCamera implements Camera, Intercom
         if (this.hasSmartDetection)
             interfaces.push(ScryptedInterface.ObjectDetector);
 
-        if (this.hasAlarm() || this.hasSupplementLight()) 
+        if (this.hasAlarm() || this.hasSupplementLight())
             interfaces.push(ScryptedInterface.DeviceProvider);
 
+        const ptzCapabilities = await this.getPtzCapabilities();
```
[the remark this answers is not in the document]
apocaliss92: done, moved as standalone functionality run only on init of the mixin

plugins/hikvision/src/main.ts
apocaliss92: reverted

plugins/hikvision/src/main.ts
```diff
@@ -478,12 +488,76 @@ export class HikvisionCamera extends RtspSmartCamera implements Camera, Intercom
         if (this.hasSmartDetection)
             interfaces.push(ScryptedInterface.ObjectDetector);
 
-        if (this.hasAlarm() || this.hasSupplementLight()) 
+        if (this.hasAlarm() || this.hasSupplementLight())
```
koush, line 491: ugh I'm just now noticing these are async calls that will always return true because they're awaited. the previous pull request was also a mistake.

refer to the implementation of hasSmartDetection that caches the capability state.

The issue with this change is that it will flap interfaces during plugin reload since hasPtz starts out as false.

plugins/hikvision/src/main.ts
```diff
@@ -478,13 +504,92 @@ export class HikvisionCamera extends RtspSmartCamera implements Camera, Intercom
...
+            this.updatePtzCaps(ptzCapabilities);
+            this.ptzPresets = cameraPresets;
+        } catch (e) {
+            if ((e.statusCode && e.statusCode === 403) ||
+                (typeof e.message === 'string' && e.message.includes('403'))) {
+            } else {
+                this.console.error('Error in getPtzCapabilities', e);
+            }
+        } finally {
+            const hasPtz = !!ptzCapabilities.length;
```
koush, line 561: what will happen if the device has ptz, but at plugin reload the device is disconnected? I think ptz will disappear?

koush: I think if ptz is enabled, it should remain enabled until it can be definitively be determined that ptz is not available (which is a scenario that shouldn't ever happen really).

This will guard against issues with connectivity during plugin start.

koush: alternatively rather than using auto detection for this state, could rely entirely on user input as a setting? then it would only need to pull the presets and cache them on startup/enable.

apocaliss92: done in a way to never disable it if it gets detected once, same for the capabilities

akifbayram: image

Settings concept from earlier (ptz instead of doorbell)

apocaliss92: Ptz would need 3 specific buttons for tilt pan and zoom

apocaliss92: image

apocaliss92: We can go for this if you guys prefer

PR #1802, opened 2025-05-07
feat(plugins/tensorflow-lite): python 3.11 support

dtomvan:
[Python 3.9 will be EOL sooner rather than later](https://github.com/google-coral/pycoral/issues/156).

This PR replaces `tflite-runtime` with `ai-edge-litert`, and `pycoral` with
`tflite-support` (because `pycoral` depends on `tflite-runtime`). The
`tflite-runtime` and `pycoral` have not been released for four years now, and
they can only be used with Python 3.9, for which support will start dropping
left and right. `tflite-support` is now the bottleneck, since it only supports
Python `3.11`, but that may change as it's not dropped yet by Google ([1](https://developers.googleblog.com/en/tensorflow-lite-is-now-litert/) and [2](https://ai.google.dev/edge/litert/libraries/task_library/object_detector#run_inference_in_python).
I might iterate on this further to use [mediapipe](https://ai.google.dev/edge/mediapipe/solutions/vision/object_detector/python#image) instead.

One caveat: I am not aware of any way to bring use Edge TPU's without
`pycoral`, or under any Python version newer than 3.9 at all, so for the time
being I have dropped that.

### DIFF
--- a/plugins/tensorflow-lite/package-lock.json
+++ b/plugins/tensorflow-lite/package-lock.json
@@ -13,7 +13,7 @@
       },
       "../../sdk": {
          "name": "@scrypted/sdk",
-         "version": "0.3.116",
+         "version": "0.5.12",
          "dev": true,
          "license": "ISC",
          "dependencies": {
--- a/plugins/tensorflow-lite/package.json
+++ b/plugins/tensorflow-lite/package.json
@@ -35,12 +35,12 @@
       ],
       "runtime": "python",
       "pythonVersion": {
-         "default": "3.9",
+         "default": "3.11",
          "darwin": {
-            "arm64": "3.10"
+            "arm64": "3.11"
          },
          "win32": {
-            "x64": "3.9"
+            "x64": "3.11"
          }
       },
       "type": "API",
--- a/plugins/tensorflow-lite/src/requirements.txt
+++ b/plugins/tensorflow-lite/src/requirements.txt
@@ -1,6 +1,5 @@
---extra-index-url https://google-coral.github.io/py-repo/
 # must ensure numpy is pinned to prevent dependencies with an unpinned numpy from pulling numpy>=2.0.
-numpy==1.26.4
-pycoral~=2.0
-tflite-runtime==2.5.0.post1
-Pillow==10.3.0
+numpy>=1.17.0,<2.0.0
+ai-edge-litert==1.2.0
+pillow==10.3.0
+tflite-support>=0.4.4,<0.5.0
--- a/plugins/tensorflow-lite/src/tflite/__init__.py
+++ b/plugins/tensorflow-lite/src/tflite/__init__.py
@@ -4,26 +4,18 @@ import json
 import threading
 
 from PIL import Image
-from pycoral.adapters import detect
 
 from .tflite_common import *
 
-loaded_py_coral = False
-try:
-    from pycoral.utils.edgetpu import list_edge_tpus, make_interpreter
+from tflite_support.task import vision, core, processor
 
-    loaded_py_coral = True
-    print("coral edge tpu library loaded successfully")
-except Exception as e:
-    print("coral edge tpu library load failed", e)
-    pass
 import asyncio
 import concurrent.futures
 import re
 from typing import Any, Tuple
 
 import scrypted_sdk
-import tflite_runtime.interpreter as tflite
+from ai_edge_litert.interpreter import Interpreter
 from .yolo_separate_outputs import *
 from scrypted_sdk.types import Setting, SettingValue
 
@@ -74,17 +66,6 @@ class TensorFlowLitePlugin(
     def __init__(self, nativeId: str | None = None, forked: bool = False):
         super().__init__(nativeId=nativeId, forked=forked)
 
-        edge_tpus = None
-        try:
-            edge_tpus = list_edge_tpus()
-            print("edge tpus", edge_tpus)
-            if not len(edge_tpus):
-                raise Exception("no edge tpu found")
-        except Exception as e:
-            print("unable to use Coral Edge TPU", e)
-            edge_tpus = None
-            pass
-
         model_version = "v14"
         model = self.storage.getItem("model") or "Default"
         if model not in availableModels:
@@ -92,6 +73,7 @@ class TensorFlowLitePlugin(
             model = "Default"
         defaultModel = model == "Default"
         branch = "main"
+        suffix = ""
 
         labelsFile = None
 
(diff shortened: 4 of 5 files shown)


koush:
does this work on mad and windows too?

python being EOL isn't an issue as the plugin will download the necessary python automatically.

koush:
to be honest, I consider coral as a whole to be deprecated since its quite dated hardware at this point, I'm hesitant to introduce changes to this plugin that may break existing users. those users can ride it out until they replace coral.

dtomvan:
> does this work on Mac and windows too?

I would have no idea. Seems like these libraries by Google are mostly platform-agnostic, as they provide native [wheels](https://pypi.org/project/ai-edge-litert/#files), but I don't think they support Windows as it isn't mentioned [in the docs](https://ai.google.dev/edge/litert). LiteRT should be compatible with its predecessor `tflite-runtime`, except for that one line where the way you initialize an `Interpreter` has changed. It seems to initialize correctly here on my NixOS system, and I think it would work just fine on MacOS.

> python being EOL isn't an issue as the plugin will download the necessary python automatically.

Well, I know it tries to download [portable python](https://github.com/bjia56/portable-python/releases/latest), but they aren't as portable as you think. When the user wants to specify their own python installation, they do so through `SCRYPTED_PYTHON_PATH` or `SCRYPTED_PYTHON3xx_PATH`. Restricting the python version to `>=3.5,<=3.9` only makes it difficult to find a package for your linux distro (or [homebrew](https://formulae.brew.sh/formula/python@3.9) for that matter). Especially on non-FHS distro's such as NixOS (which is why this old python requirement came to my attention in the first place) you can't just "download a binary and expect it to work", because if it links with some library in `/usr/lib/libz.so` for example, it will not find it.

For example:
```
$ grep PRETTY_NAME /etc/os-release
PRETTY_NAME="NixOS 25.05 (Warbler)"

$ wget https://github.com/bjia56/portable-python/releases/download/cpython-v3.9.22-build.0/python-full-3.9.22-linux-x86_64.zip
[...]

$ unzip -D python-full-3.9.22-linux-x86_64.zip
[...]

$ ./python-full-3.9.22-linux-x86_64/bin/python
bash: ./python-full-3.9.22-linux-x86_64/bin/python: cannot execute: required file not found
```

Which means that, since Python 3.9 will get removed from NixOS in the upcoming May 2025 release, I will not be able to install `@scrypted/tensorflow-lite` anymore from the admin panel because I cannot set `SCRYPTED_PYTHON39_PATH` to anything appropriate anymore without packaging Python 3.9 myself.

bjia56:
What is the binary dependency that portable python has which prevents running on NixOS? I tried to build it so that the only host dependency is glibc, which I assume is not available for NixOS?

dtomvan:
If I really wanted, I could package `bjia56/portable-python` from the Github releases at version `3.9.22` with:
```nix
{
  lib,
  stdenv,
  fetchzip,
  freetype,
  autoPatchelfHook,
}:
let
  version = "3.9.22";
  buildNumber = "build.0";
  majorMinor = lib.concatStrings (lib.take 2 (lib.splitVersion version));

  suffix =
    rec {
      aarch64-linux = "linux-aarch64";
      x86_64-linux = "linux-x86_64";
      aarch64-darwin = "darwin-universal2";
      x86_64-darwin = aarch64-darwin;
    }
    ."${stdenv.hostPlatform.system}";
in
stdenv.mkDerivation {
  pname = "portable-python${majorMinor}-bin";
  version = "${version}-${buildNumber}";

  src = fetchzip {
    url = "https://github.com/bjia56/portable-python/releases/download/cpython-v${version}-${buildNumber}/python-full-${version}-${suffix}.zip";
    hash =
      rec {
        aarch64-linux = "sha256-7qw0+dKRmLvTC5p7GTJgjBrLsJsu6hvZp5zIIXAes9A=";
        x86_64-linux = "sha256-hPItLZ5JwWV0Lldf1nUT7KqTMmMQ2eH7LqZSXyEpHc8=";
        aarch64-darwin = "sha256-DyNDa68uWnOlqzo0JXuR6/SW9xN+8BSr15p2TLM/deo=";
        x86_64-darwin = aarch64-darwin;
      }
      ."${stdenv.hostPlatform.system}";
  };

  buildInputs = [ freetype ];
  nativeBuildInputs = [ autoPatchelfHook ];

  dontBuild = true;

  installPhase = ''
    runHook preInstall

    mkdir -p $out
    cp -r * $out

    runHook postInstall
  '';

  meta = {
    description = "Portable Python binaries";
    homepage = "https://github.com/bjia56/portable-python";
    license = lib.licenses.asl20;
    mainProgram = "python3";
    platforms = [
      "aarch64-linux"
      "x86_64-linux"
      "aarch64-darwin"
      "x86_64-darwin"
    ];
  };
}
```

As you can see `libpython3.so` depends on `libfreetype`, so that also needs to be added as a dependency. Then `autoPatchelfHook` takes care of the dirty work and points all executables and DLLs to the correct places.

But this is a very hacky solution, as there already exist packages for python in NixOS which are built from source, which can be easily used with Scrypted (as long as Scrypted only uses versions which aren't too far out of date). And also this doesn't really concern you, this is a [nixpkgs](https://github.com/nixos/nixpkgs) thing. But this is the reason I wanted to submit this PR to you, because I am currently in the process [of packaging this project for Nix](https://github.com/NixOS/nixpkgs/pull/404971), which I'd rather do against a python newer than 3.9, because otherwise the package will break within a month at most.

bjia56:
I'm inspecting the Linux portable 3.9.22 for x86_64 and I don't see freetype as a dependency; in the build, it is specifically statically linked into libpython3.9.so. Is it an issue because of MacOS?

bjia56:
My opinion on this is that if there's an incompatibility that prevents portable python from working in the NixOS environment, then I can look into resolving it without needing to package it under Nix. While tflite is the most popular case and relies on an outdated version, current Scrypted plugin machinery allows any arbitrary plugin to pin to a particular python version (such as [Arlo](https://github.com/scryptedapp/arlo/blob/main/package.json#L37)) which would cause this to become a recurring issue as system python releases become EOL.

dtomvan:
> Is it an issue because of MacOS?

Happens also on Linux:
```ShellSession
$ ldd python-full-3.9.22-linux-x86_64/lib/libpython3.so
ldd: warning: you do not have execution permission for `python-full-3.9.22-linux-x86_64/lib/libpython3.so'
        linux-vdso.so.1 (0x00007f2c67605000)
        libfreetype.so.6 => not found
        libpython3.9.so => not found
```

> current Scrypted plugin machinery allows any arbitrary plugin to pin to a particular python version (such as [Arlo](https://github.com/scryptedapp/arlo/blob/main/package.json#L37)) which would cause this to become a recurring issue as system python releases become EOL.

I think it's good practice for these plugins to keep up with modern Python versions, because after EOL you also don't get security backports anymore, and your app might become vulnerable.

> My opinion on this is that if there's an incompatibility that prevents portable python from working in the NixOS environment, then I can look into resolving it without needing to package it under Nix

The best solution would be to use python that's already present and maintained under nixpkgs IMHO. But if you really want to use portable python for NixOS dynamically as you do right now, you could add a hook that:
- Checks wether we are on NixOS right now
- If so, after installing Portable Python, run `nix-shell -p auto-patchelf freetype pkg-config --run 'auto-patchelf --paths lib bin --libs `pkg-config --variable libdir freetype2`'` in the directory of the python installation.

Then the portable python installation will be modified in-place to work with NixOS.

bjia56:
Ok, I see the issue - libpython3.so depends on freetype but libpython3.9.so does not. This is fixable by tweaking the portable build, I will look into it in the coming days. Not entirely sure why it shows up on Nix and not on Ubuntu since normal python invocation only loads libpython3.9.so and doesn't touch libpython3.so, but perhaps a dependent package wants libpython3.

koush:
> I think it's good practice for these plugins to keep up with modern Python versions, because after EOL you also don't get security backports anymore, and your app might become vulnerable.

Sure, if the upstream dependencies are as well. Building tflite is a nightmare and this is to extend support for devices that are 6 years old now. The juice isn't worth the squeeze here. The edgetpus can't fit any of the modern models on them either (face recognition and lpr).

> Which means that, since Python 3.9 will get removed from NixOS in the upcoming May 2025 release, I will not be able to install `@scrypted/tensorflow-lite` anymore from the admin panel because I cannot set `SCRYPTED_PYTHON39_PATH` to anything appropriate anymore without packaging Python 3.9 myself.

Docker/lxc is a a requirement for tflite support specifically due to these difficulties.

dtomvan:
> This is fixable by tweaking the portable build,

@bjia56 I don't know wether that will make it work on NixOS or not. I will give one of the artifacts a try without any tweaks, see if that helps. Thank you for the quick response.

> Sure, if the upstream dependencies are maintained or at least easily buildable

Yes, I get that. I think that the old dependencies have been superseded by the ones in my patch, and hence reason enough to move on right? I guess unless someone uses Windows and still wants to rely on this old hardware...

> The juice isn't worth the squeeze here

Did you mean that you want to stop publishing new versions of this plugin or do you mean that you don't have intention of changing the Python requirement?

> Docker/lxc is a a requirement for tflite support specifically due to these difficulties.

Ah, that makes sense. So I suppose that's why you don't mind staying on a version which might become less and less available?

dtomvan:
Alright. Seems fair. Thanks for clarification. I'll close this then. At least we found a way to make things more compatible: after that freetype static-link change by @bjia56 it will probably become possible to use this plugin on NixOS, as the Python interpreter doesn't have to come from Nixpkgs anymore. So I think I don't even need to keep my own fork. Thanks for helping me out.

PR #1804, opened 2025-05-09
Fix events stopping for NVRs

apocaliss92:


apocaliss92:
Fix confirmed by an affected user

PR #1808, opened 2025-05-20
Update bticino-camera.ts

mrMiimo:
Extended the doorbell motion timeout from 20s to 60s. HomeKit Secure Video (HKSV) clips were limited to ~10s due to the short event duration. By keeping the motion event active longer, HKSV can record longer clips if activity persists.

### DIFF
--- a/plugins/bticino/src/bticino-camera.ts
+++ b/plugins/bticino/src/bticino-camera.ts
@@ -642,8 +642,11 @@ export class BticinoSipCamera extends ScryptedDeviceBase implements MotionSensor
             this.binaryState = true
             setTimeout( () => {
                 // Assumption that flexisip only holds this call active for 20 seconds ... might be revised
+                // Extended the doorbell motion timeout from 20s to 60s.
+                // HomeKit Secure Video (HKSV) clips were limited to ~10s due to the short event duration.
+                // By keeping the motion event active longer, HKSV can record longer clips if activity persists.
                 this.reset()
-            }, 20 * 1000 )            
+            }, 60 * 1000 )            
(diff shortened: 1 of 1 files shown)


koush:
@slyoldfox

slyoldfox:
@mrMiimo did you build and deploy a patched version for you locally and can you confirm it actually works (because I'm running a whole other version of the plugin locally atm and I can't test it much myself).

mrMiimo:
@slyoldfox  I wasn’t able to set up a full development environment, but considering the minimal scope of the change (just a value), I thought it wasn’t worth the overhead.
I’ll test it right away and, if needed, I’ll report back immediately so you can roll it back.

slyoldfox:
> @slyoldfox I wasn’t able to set up a full development environment, but considering the minimal scope of the change (just a value), I thought it wasn’t worth the overhead. I’ll test it right away and, if needed, I’ll report back immediately so you can roll it back.

Just a bit of context, the value is choosen because the unit itself hangs up the connection around 30 seconds. So even though `reset()` only resets some variables, I was trying to keep it more or less aligned with that timing.

I also need to point out that the `reset()` probably will be happening earlier as well, since the main entry point is in the invite handler on https://github.com/koush/scrypted/blob/main/plugins/bticino/src/bticino-inviteHandler.ts#L27 - you will find the same timeout there.

I think at some point I did try to increase the values and I'm not sure that I ever got HKSV recording more than 10 seconds working (not sure where it goes wrong tbh). So please do fiddle around with it and test it out with an actual doorbell press on a local instance before we just merge this.

PR #1809, opened 2025-05-20
Update bticino-camera.ts

mrMiimo:
keyframes every 2 seconds to improve HKSV compatibility


mrMiimo:
> cameras should not transcode by default as this does. it will tax the system. This can be done using the synthetic streams option.

Thanks! You're right, I’ll avoid transcoding by default.
In my case, the device uses Speex (non-updatable), so I need to transcode audio to AAC for HKSV to work.

PR #1841, opened 2025-06-29
Fallback pip to latin

VitorFurlanetti:
Had an issue running Scrypted in Windows under a PT-BR install. The default user on this case is created as "Usuário" and pip was complaining about not finding the home dir. 
With this fix I'm able to run it in Windows.


koush:
I think this should catch the error if the second decode fails too, as this is a catastrophic error, not logging the pip output is acceptable in case it can not be decoded.

koush:
oh I guess it can't fail

PR #1936, opened 2025-11-28
Attempt POST devinfo

apocaliss92:
This PR attempts to fetch the DevInfo API with a POST when the GET fails (same approach as for GetAbility). Probably newer models do not support the GET

This should be merged after https://github.com/koush/scrypted/pull/1935/files and released as beta

### DIFF
--- a/plugins/reolink/src/reolink-api.ts
+++ b/plugins/reolink/src/reolink-api.ts
@@ -294,28 +294,20 @@ export class ReolinkCameraClient {
     async jpegSnapshot(timeout = 10000) {
         const url = new URL(`http://${this.host}/cgi-bin/api.cgi`);
 
-        const body = [
-            {
-                cmd: "Snap",
-                channel: this.channelId.toString(),
-                rs: Date.now().toString()
-            }
-        ];
+        const params = url.searchParams;
+        params.set('cmd', 'Snap');
+        params.set('channel', this.channelId.toString());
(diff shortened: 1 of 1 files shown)


plugins/reolink/src/reolink-api.ts
```diff
@@ -280,7 +280,7 @@ export class ReolinkCameraClient {
 
             error = response.body?.[0]?.error;
             if (error) {
-                this.console.error('error during call to getAbility GET, Trying with POST', error);
```
apocaliss92: Fixed here a typo of the error log, it is a POST and not a GET (copy paste error)

koush:
I think maybe just switch it to only POST since that's what their API indicates, and if it breaks something we can switch to use the fallback method. This seems to be an implementation error on our part that happened to work due to luck.

The other stuff should be in separate pull requests.

apocaliss92:
Sure I'll do later, what should be in a separate pr? Besides a log text it's 2 different methods with the same fallback post fix (snapshot and devInfo)

koush:
the device is home hub check fix should be separate

apocaliss92:
That's the commit already merged, I have to rebase

koush:
oh ok

apocaliss92:
Not necessary anymore

koush:
why's that?

PR #1938, opened 2025-12-02
Add pluginId to reolink/hikvision cameras interface

apocaliss92:


koush:
Use pluginid property

apocaliss92:
Is it available in the device filter and canMixin? That's the only reason I would need these

PR #1939, opened 2025-12-03
sdk: add WebComponent

bjia56:


koush:
can be done via package.json scrypted key as discussed on discord https://github.com/scryptedapp/llm/blob/db4baa84341169195a7819fc14d03a497b0b13eb/package.json#L40

PR #1947, opened 2025-12-26
Reolink nvr

apocaliss92:
Adds explicit support to NVR devices. User can add an hub/NVR device as standalone device, the device will discover and allow adopting all the connected devices.
It was not possibile to reuse client or camera devices since the communication is completely different, since there is now only 1 authentication session running for all the devices, NVR devices poll data for all the connected devices.
Currently in testing on my home setup with an home hub and 4 connected battery devices

The changes in the main should be ONLY to fork the creation of the device if the boolean is set during the creation and extend the getDevice too

### DIFF
--- a/plugins/reolink/src/nvr/api.ts
+++ b/plugins/reolink/src/nvr/api.ts
@@ -116,7 +116,7 @@ export class ReolinkNvrClient {
     loggingIn = false;
     loggedIn = false;
     rebooting = false;
-    conmnectionTime = Date.now();
+    connectionTime = Date.now();
     console: Console;
     host: string;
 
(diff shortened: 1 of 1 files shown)


plugins/reolink/src/main.ts
```diff
@@ -1149,63 +1153,119 @@ class ReolinkProvider extends RtspProvider {
...
-            catch (e) {
-                this.console.error('Error adding Reolink camera', e);
-                throw e;
+        const ip = settings.ip?.toString();
+        const httpPort = settings.httpPort;
+        const rtspPort = settings.rtspPort;
+        const isNvr = settings.isNvr?.toString() === 'true';
+        const httpAddress = `${ip}:${httpPort || 80}`;
+
+        if (isNvr) {
```
koush: lets move this entire block into a `return createNvrDevice`. Then remove the else block so the original code doesn't need to be branched.

PR #1953, opened 2025-12-31
python: use create_task instead of run_coroutine_threadsafe

raman325:
## Summary

- Replace `asyncio.run_coroutine_threadsafe()` with `asyncio.create_task()` in the RPC read loop

## Problem

`run_coroutine_threadsafe` is designed for scheduling coroutines from a **different thread** onto the event loop. However, `readLoop` is already an async function running on the event loop, so using `run_coroutine_threadsafe` adds unnecessary thread-safe queue overhead for every RPC message.

## Solution

Use `create_task()` which is the native and more efficient way to spawn concurrent coroutines when already running on the event loop.

## Test plan

- [ ] Verify Python plugins continue to work correctly
- [ ] No functional change expected, only minor performance improvement

### DIFF
--- a/server/python/rpc_reader.py
+++ b/server/python/rpc_reader.py
@@ -206,7 +206,7 @@ class RpcConnectionTransport(RpcTransport):
         return self.writeMessage(bytes(buffer), reject)
 
 
-async def readLoop(loop, peer: rpc.RpcPeer, rpcTransport: RpcTransport):
+async def readLoop(peer: rpc.RpcPeer, rpcTransport: RpcTransport):
     deserializationContext = {"buffers": []}
 
     while True:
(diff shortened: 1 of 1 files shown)


koush:
was this tested with a server build

raman325:
not yet these were meant to be opened in draft. Will be testing them before opening them up

raman325:
The ONNX Python plugin loaded and functioned correctly:
  - ✅ RPC communication between Node.js server and Python plugin works
  - ✅ Downloaded and loaded 6 ONNX models (yolov9, face detection, inception, craft, vgg, CLIP)
  - ✅ CoreML execution provider initialized successfully
  - ✅ No RPC-related errors or performance issues

PR #1954, opened 2025-12-31
python: use asyncio.Lock instead of threading.Lock for RPC send

raman325:
## Summary

- Replace `threading.Lock` with `asyncio.Lock` for the RPC send mutex
- Wrap send operations in `asyncio.create_task()` to maintain the existing sync interface

## Problem

`threading.Lock` blocks the **entire event loop** while waiting to acquire the lock. When there's contention on the send lock, all other coroutines stall until the lock is released.

```python
# Before: blocks event loop
mutex = threading.Lock()
def send(message, ...):
    with mutex:  # ← All coroutines blocked here
        ...
```

## Solution

Use `asyncio.Lock` which yields control back to the event loop while waiting, allowing other work to continue:

```python
# After: yields to event loop
send_lock = asyncio.Lock()
async def async_send(message, ...):
    async with send_lock:  # ← Other coroutines can run
        ...

def send(message, ...):
    asyncio.create_task(async_send(...))
```

## Test plan

- [ ] Verify Python plugins continue to work correctly
- [ ] Test with concurrent RPC operations to verify no deadlocks
- [ ] Expected improvement under concurrent message sending

### DIFF
--- a/server/python/rpc_reader.py
+++ b/server/python/rpc_reader.py
@@ -7,7 +7,6 @@ import json
 import multiprocessing.connection
 import os
 import pickle
-import threading
 from asyncio.events import AbstractEventLoop
 from concurrent.futures import ThreadPoolExecutor
 from typing import List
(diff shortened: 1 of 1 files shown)


koush:
did you test this on a full server build

raman325:
The ONNX Python plugin loaded and functioned correctly:
  - ✅ RPC communication between Node.js server and Python plugin works
  - ✅ Downloaded and loaded 6 ONNX models (yolov9, face detection, inception, craft, vgg, CLIP)
  - ✅ CoreML execution provider initialized successfully
  - ✅ No RPC-related errors or deadlocks

koush:
send needs to remain sync, and I dont feel comfortable having this slop coded tbh. I also dont see a pressing need. I'll consider doing it myself but this is a critical path that I dont want changed willy nilly.

koush:
asyncio.Lock is not thread safe anyways. It's only meant to be called from the event loop that created it. send may be called from **any thread** because the caller may be calling from any thread, including one that may not even have a loop. changing it to async_send also doesn't change/fix this.

Furthermore, scheduling on the event loop thread is inherently uses synchronization primitives under the hood because the loop's queue needs to be locked to add something to it. Arguably one could use run_coroutine_threadsafe (which also uses a mutex internally) but I don't really want to change something that isn't broken.

koush:
https://chatgpt.com/share/696947a2-b9ac-800c-89f3-09c2618e101f

I don't know much about python but I am deeply familiar with how event loops work in regards to synchronization and threading, having written one for scratch for java/kotlin some time ago for another networking project/library. The chatgpt conversation is just a python specific confirmation of why I was skeptical of this change.

raman325:
fair enough! Seemed innocuous to me but I am less familiar with event loops in general

PR #1961, opened 2026-01-04
Update Google Home plugin README with disclaimer about app

jguerin:
Added a note about video streaming limitations in Google Home, based off comments made by @koush here:
https://www.reddit.com/r/Scrypted/comments/1nwrcit/unifi_protect_with_google_home/


koush:
should make this change in https://docs.scrypted.app, slowly moving all the documentation there and the plugin read me will reference that, as this would require a plugin update.

PR #1965, opened 2026-01-12
hikvision: add controls for supplemental light, audible alarm, and strobe light 

akifbayram:
Add controls for supplemental light, audible alarm, and strobe light. Including supp light mode selection, brightness adjustments, audible alarm selection and volume, strobe light mode, and state retrieval upon startup. Conditional readme.

## Smart Hybrid Camera Supplemental Light Controls
Screenshot 2026-01-12 140422

## White Light Only Camera Supplemental Light Controls
image

## Alarm Linkages
Screenshot 2026-01-12 140402

### Audio Alarm Controls
Screenshot 2026-01-12 140410

### Strobe Light Controls
image

### DIFF
--- a/plugins/hikvision/src/supplemental-light.ts
+++ b/plugins/hikvision/src/supplemental-light.ts
@@ -2,11 +2,28 @@ import { Brightness, OnOff, Readme, ScryptedDeviceBase, Setting, Settings, Setti
 import { StorageSettings } from "@scrypted/sdk/storage-settings";
 import type { HikvisionCamera } from "./main";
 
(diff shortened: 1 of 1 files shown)


koush:
I don't really want to this plugin (or other camera plugins) to become an alternate/duplicate camera portal settings. It will be a never ending bug farm. For example, this code does not retrieve the initial state correctly.

While I think there is value in being able to control the lights (turn them on/off etc, so they can get synced into homekit), I think it should be left to the user to refer to the manufacturer provided interface to do the one time configuration of the other settings themselves.

PR #1968, opened 2026-01-22
hash-based paths for model downloads

omares:
## Summary

- Replace `self.id` (runtime device ID) with hash-based paths for model downloads
- Path format: `{plugin_suffix}/{sha256(url)[:12]}/{filename}`
- Ensures consistent, reproducible, collision-free paths across plugin reinstalls

## Problem

The detection plugins used `self.id` (a Scrypted-assigned device ID) in file download paths. This ID is non-deterministic - it's assigned at runtime and changes if the plugin is reinstalled. This breaks reproducibility and wastes disk space on reinstalls.

## Solution

Use a deterministic path based on:
1. Plugin suffix extracted from `self.pluginId` (e.g., `openvino` from `@scrypted/openvino`)
2. First 12 chars of SHA256 hash of the download URL (collision-free, deterministic)

## Affected Plugins

All 6 detection plugins via symlinks: openvino, onnx, ncnn, tensorflow-lite, rknn, coreml

### DIFF
--- a/plugins/openvino/src/predict/clip.py
+++ b/plugins/openvino/src/predict/clip.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import base64
+import hashlib
 from typing import Tuple
 
 import scrypted_sdk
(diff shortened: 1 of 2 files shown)


koush:
the clip model download is also moving to hugging face when I get around to it and using the same downloader as the other models, this code won't survive that change.

koush:
I believe I told you that the hugging face change is coming on discord

omares:
> I believe I told you that the hugging face change is coming on discord

Not directly; it was mentioned that the Hugging Face downloader is in use, but not that the plan is to replace the clip and custom_detect implementations.

Even better that the download approach will be consistent throughout the codebase.

PR #1995, opened 2026-03-02
onvif: fix motion stop signal not clearing motion badge

themactep:
## Problem

Cameras that correctly send `IsMotion=false` ONVIF stop signals (e.g. thingino-based cameras) still keep the motion badge lit for ~30 seconds after motion stops.

Two bugs combine to cause this:

### 1. `RuleEngine/CellMotionDetector/Motion` stop is silently dropped (`onvif-api.ts`)

```typescript
else if (eventTopic.includes('RuleEngine/CellMotionDetector/Motion')) {
    // unclear if the IsMotion false is indicative of motion stop?
    if (event.message.message.data.simpleItem.$.Name === 'IsMotion' && dataValue) {
        ret.emit('event', OnvifEvent.MotionBuggy);  // only fires on true
    }
    // IsMotion=false → nothing emitted
}
```

The ONVIF spec defines `IsMotion=false` in a `CellMotionDetector/Motion` message as an explicit motion-end signal — the same semantics as `VideoSource/MotionAlarm` with `IsMotion=false`, which is already handled correctly. The comment "unclear if..." is no longer warranted; this has been consistent across ONVIF-compliant cameras.

### 2. `MotionStop` resets the 30-second debounce timer instead of clearing motion (`onvif-events.ts`)

```typescript
else if (event === OnvifEvent.MotionStop) {
    if (thisDevice.motionDetected)
        triggerMotion();  // ← resets 30s timer, keeps badge lit
    // thisDevice.motionDetected = false;  ← commented out
}
```

`triggerMotion()` always sets `motionDetected = true` and arms a fresh 30-second timeout. So receiving a stop signal actually *prevents* the badge from clearing until the full 30s elapses.

## Fix

1. **`onvif-api.ts`**: Treat `CellMotionDetector/Motion` `IsMotion=false` as `MotionStop`, consistent with `VideoSource/MotionAlarm`.

2. **`onvif-events.ts`**: On `MotionStop`, immediately clear `motionDetected` and cancel the debounce timer. The 30-second timeout-based debounce (`triggerMotion`) is preserved for `MotionStart` and `MotionBuggy`, which still covers cameras that never send stop events.

## Backward compatibility

- Cameras that **do** send stop events: motion badge clears immediately ✓  
- Cameras that **don't** send stop events: unchanged — `MotionStart` still uses the 30-second `triggerMotion` debounce ✓  
- `MotionBuggy` path is unchanged ✓

### DIFF
--- a/plugins/coreml/package-lock.json
+++ b/plugins/coreml/package-lock.json
@@ -1,12 +1,12 @@
 {
    "name": "@scrypted/coreml",
-   "version": "0.1.91",
+   "version": "0.1.92",
    "lockfileVersion": 2,
    "requires": true,
    "packages": {
       "": {
          "name": "@scrypted/coreml",
-         "version": "0.1.91",
+         "version": "0.1.92",
          "devDependencies": {
             "@scrypted/sdk": "file:../../sdk"
          }
--- a/plugins/coreml/package.json
+++ b/plugins/coreml/package.json
@@ -50,5 +50,5 @@
    "devDependencies": {
       "@scrypted/sdk": "file:../../sdk"
    },
-   "version": "0.1.91"
+   "version": "0.1.92"
 }
--- a/plugins/coreml/src/requirements.txt
+++ b/plugins/coreml/src/requirements.txt
@@ -2,4 +2,5 @@ coremltools==8.0
 Pillow==10.3.0
 opencv-python-headless==4.10.0.84
 
-transformers==4.52.4
+transformers==5.2.0
+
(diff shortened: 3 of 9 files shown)


koush:
The issue with this change is that all the cameras in scrypted debounce the motion start/stop events to normalize it for downstream. Sending stop events and then a start event within a few seconds will cause HKSV to stop recording and then immediately start recording. It basically acts as a filter on noisy motion sensors. The timeout on the motion stop should remain.

PR #2002, opened 2026-03-23
wyze: add include/exclude camera filter setting

ropelletier:
Adds two new plugin-level settings to the Wyze plugin:

- Camera Filter Mode: dropdown with options Disabled / Include Only / Exclude
- Camera Filter: multi-select list populated dynamically from all discovered cameras, allowing the user to pick which cameras to include or exclude

When filter mode is "Include Only", only cameras whose nicknames appear in the filter list are registered with Scrypted. When "Exclude", all cameras except those in the list are registered. "Disabled" (default) preserves existing behaviour — all cameras are registered.

All discovered cameras are stored in self.cameras before filtering so the multi-select choices are always fully populated in the settings UI, even when a filter is active.

### DIFF
--- a/plugins/wyze/src/main.py
+++ b/plugins/wyze/src/main.py
@@ -1113,11 +1113,22 @@ class WyzePlugin(scrypted_sdk.ScryptedDeviceBase, DeviceProvider):
         self.authInfo = auth_info
         self.account = wyzecam.get_user_info(auth_info)
         cameras = wyzecam.get_camera_list(auth_info)
-        # await self.pollEvents()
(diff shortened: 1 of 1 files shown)


koush:
scrypted does not filter devices that are provided via a sync. you can disable them by turning off all extensions and/or changing their type. you can also prevent non-admin users from seeing them. they become data only, and no streams are initiated.

ropelletier:
OK, but isn't this much easier and more intuitive. This is one step, your solution is multiple steps and cumbersome. Maybe I am missing something?

koush:
yeah you can change the type to unknown, thats one step.

PR #2017, opened 2026-04-19
feat: implement momentary switch auto-lock delay for HomeKit relays

t09:
Added an 'Auto-Lock Delay' configuration option to the HomeKit mixin. This completely decouples HomeKit from Scrypted's internal cached 'Unlocked' state when dealing with stateless door strike relays (like Dahua) to prevent the Home app UI from getting permanently stuck on 'Unlocking...'."


koush:
Should probably be fixed in the amcrest plugin to fake the state to be whatever the user sets it to.

PR #2031, opened 2026-05-06
NVR: Add external fork-spawn watchdog to prevent host OOM cascade

amirandap:
## Overview
Adds an external watchdog (Python + systemd service) that monitors Scrypted NVR fork-spawn rates and applies circuit-breaker protection to prevent host-level OOM cascades on multi-camera deployments.

## Problem
On systems with 6+ cameras using Intel iGPU (VAAPI hwaccel), the NVR plugin can enter a fork-spawn storm triggered by:
1. Camera with NVR Object Detection mixin enabled **without a license** → 334 forks in 10 min
2. VAAPI decoder crashes during i915 GuC reset under load → immediate respawn with no backoff
3. Cascading failures: container OOM → cgroup kill → host OOM → QEMU/HA VM killed → bootloop

Observed: **87–109 fork spawns / 5–10 min** → **12 GiB memory usage** → **host crash**

## Solution
External circuit breaker that tails `docker logs -f scrypted`:
- **Per-camera limit**: > 30 spawns / 60 s → warning log + Home Assistant notification
- **Aggregate limit**: > 150 spawns / 60 s → controlled `docker restart scrypted` (with 5 min cooldown)

This prevents the spawn storm from cascading to host-level OOM while preserving other camera streams.

## Why external?
Patching the minified NVR plugin bundle (`main.nodejs.js`) is fragile and lost on every update. The watchdog operates Docker-externally, requires no code changes to Scrypted, and survives updates.

## Deployment
```bash
# Copy script
cp scripts/scrypted-fork-watchdog.py /root/scrypted/scripts/

# Create systemd service
cat > /etc/systemd/system/scrypted-fork-watchdog.service << 'EOF'
[Unit]
Description=Scrypted fork-spawn watchdog (circuit breaker)
After=docker.service
Requires=docker.service

[Service]
Type=simple
Environment=HA_TOKEN=<your_ha_token_here>
ExecStart=/usr/bin/python3 /root/scrypted/scripts/scrypted-fork-watchdog.py
Restart=always
RestartSec=10

[Install]
WantedBy=multi-user.target
EOF

# Enable
systemctl daemon-reload
systemctl enable --now scrypted-fork-watchdog.service
```

## Results
After deployment (verified on i5-10500, 6 cameras, Intel iGPU):
```
Fork rate: 11 / 60s (was 109 / 60s)
Memory: 1.13 GiB steady (was 12 GiB OOM)
VAAPI errors: 0 (was 165 / 5 min)
[...]

PR #2036, opened 2026-05-10
feat(server): ForkGovernor — per-plugin exit-aware backoff and circuit breaker for sdk.fork()

amirandap:
## Motivation

Plugins like `@scrypted/nvr` call `sdk.fork()` very frequently — once per motion event per camera per pipeline (recording, detection, snapshots). If a fork crashes immediately on startup (ENOENT writing event JSON, GuC iGPU submission failure, OOM), the consumer typically respawns right away. A handful of misbehaving cameras can therefore spawn **hundreds of doomed child processes per minute**, exhausting CPU, GPU contexts, file descriptors, and memory until the entire container falls over.

**Observed failure cascade on a 6-camera deployment:**
| Symptom | Count | Window |
|---------|-------|--------|
| `starting fork @scrypted/nvr` | 8 767 | ~12 hours |
| Container memory growth | 1.6 → 12 GiB | ~1 hour |
| Cgroup OOM kill | yes | repeated |
| Host OOM kill of QEMU/KVM HA VM | yes | bootloop |

This is the failure mode that motivated the external watchdog in #2031, which was correctly closed with "needs a proper fix in nvr". The proper fix is **here, in core**: enforce back-pressure on the SDK fork API itself, so every plugin (including closed-source ones like NVR) gets resilience for free.

## Design

`ForkGovernor` is a per-`(plugin, key)` rate limiter — where `key` is typically the camera/device ID — that sits in `sdk.fork()`:

1. **Sliding-window tracking** of both fork *attempts* and *crashes* (abnormal exit or exit within 5 s of spawn).
2. **Exponential backoff** before each fork when crashes accumulate: `min(30 s, 250 ms × 2^crashes)` ± 20% jitter.
3. **Hard circuit breaker**: if a plugin exceeds `HARD_LIMIT` forks **and** `HARD_CRASH_LIMIT` crashes in a 60-second window, reject the fork immediately with a clear error. NVR's existing `.catch()` path then makes the host stable.
4. **Auto-reset**: if a fork lives past 60 seconds, its crash counter is cleared — well-behaved workers are not penalized for old failures.
5. **Fully opt-out** via `SCRYPTED_FORK_GOVERNOR_DISABLE=1` env var.

### Tunables (all optional env vars)
| Var | Default | Meaning |
|-----|---------|---------|
| `SCRYPTED_FORK_GOVERNOR_DISABLE` | — | Set to `1` to disable |
| `SCRYPTED_FORK_GOVERNOR_WINDOW_MS` | 60 000 | Sliding window |
| `SCRYPTED_FORK_GOVERNOR_SOFT_LIMIT` | 30 | Forks/window before backoff |
| `SCRYPTED_FORK_GOVERNOR_HARD_LIMIT` | 90 | Forks/window before rejection |
| `SCRYPTED_FORK_GOVERNOR_HARD_CRASH_LIMIT` | 10 | Crashes/window before rejection |
| `SCRYPTED_FORK_GOVERNOR_MIN_BACKOFF_MS` | 250 | Min backoff |
| `SCRYPTED_FORK_GOVERNOR_MAX_BACKOFF_MS` | 30 000 | Max backoff cap |

## Testing

Smoke tests included in `server/test/fork-governor-smoke.ts`. Verified in production: fork rate dropped from 8 767/12 h to <20/h after the fix, container memory stabilized at ~950 MiB.

### DIFF
--- a/server/src/plugin/fork-governor.ts
+++ b/server/src/plugin/fork-governor.ts
@@ -1,0 +1,255 @@
+/**
+ * ForkGovernor — per-plugin fork-spawn rate limiter, exponential-backoff,
+ * and crash-aware circuit breaker for `scrypted.fork()`.
+ *
+ * Motivation
+ * ----------
+ * Plugins like @scrypted/nvr can call `sdk.fork()` very frequently — once per
+ * motion event per camera per pipeline (recording, detection, snapshots).
+ * If a plugin's fork crashes immediately on startup (e.g. ENOENT writing event
+ * JSON, GuC submission failure on iGPU, OOM) the consumer typically respawns
+ * right away. A handful of misbehaving cameras can therefore spawn hundreds of
+ * doomed child processes per minute, exhausting CPU, GPU contexts, FDs, and
+ * memory until the entire container falls over (host-level OOM killer, i915
+ * GuC reset cascades, etc.).
+ *
+ * This was the failure mode that motivated the external watchdog in
+ * https://github.com/koush/scrypted/pull/2031, which Koush correctly closed
+ * with "needs a proper fix in nvr". The proper fix is here, in core: enforce
+ * back-pressure on the SDK fork API itself, so every plugin (including
+ * closed-source ones like NVR) gets resilience for free.
+ *
+ * Algorithm
+ * ---------
+ * Per `(plugin, key)` — where `key` is `options.id` (usually a camera/device
+ * id) or `options.name`, falling back to the literal string "default" — we
+ * maintain:
+ *
+ *   - `attempts[]`  : timestamps of recent fork() calls (sliding window)
+ *   - `crashes[]`   : timestamps of recent abnormal exits (exit code != 0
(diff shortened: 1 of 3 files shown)


amirandap:
Closing this. After tracing the actual root cause of the fork storm (NVR disk threshold — disk was at 86% full, NVR requires >15% free to enable truncate-mode recording), the ForkGovernor approach creates more problems than it solves:

- **False positive on normal startup**: 6 cameras × 1 init fork each = 6 fast-exiting forks in <30s, identical signature to a crash storm. The governor tripped on clean startup and blocked all NVR from starting — worse than any fork storm.
- **Treats the symptom, not the cause**: Every fork storm observed had a concrete upstream root cause (disk full, ENOENT race, shmem leak). The real fixes are PRs #2034 and #2035.

PRs #2034 (mkdirSync before fallback createWriteStream) and #2035 (drain MessagePort before terminate) address the actual failure modes. With those in place, the governor is redundant.

PR #2041, opened 2026-05-12
feat: implement auto-lock delay for Dahua door strikes no.2

t09:
additional #2017:

**Amcrest plugin changes:**
- Initialize `lockState` to `Locked` on startup so HomeKit accepts the device
- `lock()` / `unlock()` now properly update `lockState` (faking the state as requested)
- Auto-lock delay: can be changed in WebUI because Amcrest not reading config of doorbell itself
- New setting "Auto-Lock Delay (Seconds)" with key `dahuaAutoLockDelay`

**HomeKit plugin changes:**
- Add `LockMechanism` service to Doorbell accessories that have the `Lock` interface
- Link it via `addLinkedService` so iOS shows the lock button in the camera view
- Use `setTimeout(150ms)` before `updateCharacteristic` to avoid HAP notification/SET-response collision - else it hangs.

The HomeKit change was necessary because without it, the Lock interface on a Doorbell device is never exposed to HomeKit regardless of the Amcrest plugin state.

### DIFF
--- a/plugins/amcrest/src/main.ts
+++ b/plugins/amcrest/src/main.ts
@@ -1,7 +1,7 @@
 import { automaticallyConfigureSettings, checkPluginNeedsAutoConfigure } from "@scrypted/common/src/autoconfigure-codecs";
 import { ffmpegLogInitialOutput } from '@scrypted/common/src/media-helpers';
 import { readLength } from "@scrypted/common/src/read-stream";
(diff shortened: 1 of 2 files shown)


koush:
use a script or something to do this, it does not need to be in a plugin.

koush:
I dont understand the amcrest side changes, this seems like a reporting issue? does the device not report lock state?

t09:
No my Dahua VTO2000A doesn't report lock state. _http://IP/cgi-bin/accessControl.cgi?action=getDoorStatus&channel=1_ gives _back Info.status=Close_ all the time which I guess most doorbells do because only electrical lock ist installed. You would need to install a seperate sensor (magnetic/reed contact) to get status. Next was I can't read timer from Dahua itself (no event) even if you can set time in DahuaUI - so I had to implant a timer in the Amcrest plugin to fake the "lock". Had to mod homekit - else the lock wouldn't show in the videostream itself but as seperate widget.

t09:
any other ideas how to implent this?

t09:
updated :)

PR #2047, opened 2026-05-21
fix(server): treat canMixin "not implemented" as passthrough to break infinite rebuild loop

amirandap:
When a mixin provider's canMixin() throws "not implemented" (plugin loading/restarting), the catch block returns error=e (truthy). ensureProxy() then merges old interfaces back, setPluginDeviceState sees a change, calls notifyPluginDeviceDescriptorChanged → rebuildMixinTable → canMixin throws again — infinite loop at ~2/s.

**Impact:** load average spikes to 19+, event loop saturated, plugin pings time out, all plugins restart in cascade. Self-sustaining once started.

**Fix:** Detect the "not implemented" message, return passthrough=true with error=undefined. Interface set stabilizes, notify is never triggered, loop stops.

**Tested live:** load dropped from 19.3 → 1.08 within 90s. All cameras stable in HomeKit.

### DIFF
--- a/server/src/plugin/plugin-device.ts
+++ b/server/src/plugin/plugin-device.ts
@@ -294,6 +294,24 @@ export class PluginDeviceProxyHandler implements PrimitiveProxyHandler<any> {
             };
         }
         catch (e) {
+            // When canMixin throws "not implemented", the mixin provider plugin is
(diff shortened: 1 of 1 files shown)


koush:
?
implement canMixin in your plugin, what is this even fixing

PR #2053, opened 2026-05-30
onvif: advanced setting to disable motion debounce

bzoel:
Add advanced setting to disable motion debounce on ONVIF devices that are well behaved when reporting motion start/stop


koush:
I think this needs to be a bit more resilient to weird onvif behavior. log the last time a motion stop event was seen. if the last "stop" was within 8 hours, set the stop timeout to 8 hours. Otherwise use the current stop timeout.

bzoel:
Yeah, marking as draft per our Discord conversation. Will mess with it a bit and push a v2.0

PR #2056, opened 2026-05-31
fix(rebroadcast): do not send getStreamLength before play in RTMP client

### LINKED ISSUE #2055: rebroadcast: Reolink RTMP connections FIN immediately because the RTMP client sends getStreamLength before play
## Summary

The custom RTMP client introduced in commit 1349bb74 sends an AMF `getStreamLength` command between `createStream` and `play`. **Reolink cameras (and likely some other RTMP servers) FIN the TCP connection ~100–150 ms after receiving `getStreamLength`**, causing the rebroadcast prebuffer to die immediately on every connection attempt and enter an endless ~5-second restart loop.

The comment on the offending block (`// Send getStreamLength then play (matching ffmpeg's order)`) is incorrect — ffmpeg's libavformat RTMP client does **not** send `getStreamLength` for live playback, only conditionally for VOD/seekable streams. `rtmpdump` does not send it either. Removing the `getStreamLength` send makes Reolink cameras stream indefinitely without disconnects.

## Affected versions

`@scrypted/prebuffer-mixin@0.10.66` (and every version since the custom RTMP implementation landed in commit `1349bb74` on 2026-01-09).

## Source location

[`plugins/prebuffer-mixin/src/rtmp-client.ts`](https://github.com/koush/scrypted/blob/main/plugins/prebuffer-mixin/src/rtmp-client.ts#L219-L242) — lines 219–242:

```ts
    // Send getStreamLength then play (matching ffmpeg's order)
    const parsedUrl = new URL(this.url);
    // Extract stream name (after /app/)
    const parts = parsedUrl.pathname.split('/');
    const streamName = parts.length > 2 ? parts.slice(2).join('/') : '';
    const playPath = streamName + parsedUrl.search;

    this.console?.log('Sending getStreamLength with path:', playPath);
    const getStreamLengthData = this.encodeAMF0Command('getStreamLength', this.transactionId++, null, playPath);
    this.sendMessage(5, 0, RtmpMessageType.COMMAND_AMF0, 0, getStreamLengthData);

    this.console?.log('Sending play command with path:', playPath);
    this.sendPlay(this.streamId, playPath);
```

## Reproduction

1. A Reolink camera that exposes RTMP (verified on a Reolink Video Doorbell WiFi `D340W`, firmware `v3.0.0.4662_2508071282`; same behaviour reported anecdotally on other Reolink RTMP-capable models).
2. Enable the camera in Scrypted and enable RTMP prebuffer (`RTMP sub.bcs` or `RTMP main.bcs`).
3. Watch the Scrypted log for the Reolink camera.

### Observed behaviour

Every prebuffer cycle:

```
[Reolink Doorbell] Starting stream()...
[Reolink Doorbell] Connecting to RTMP server at <camera>:1935
[Reolink Doorbell] Socket connected
[Reolink Doorbell] Performing handshake...
[Reolink Doorbell] Handshake complete
[Reolink Doorbell] Sending connect command...
[Reolink Doorbell] Connect command sent
[Reolink Doorbell] Server set chunk size to 4096
[Reolink Doorbell] Received _result for connect
[Reolink Doorbell] Sending createStream...
[Reolink Doorbell] Got createStream _result
[Reolink Doorbell] Sending getStreamLength with path: channel0_sub.bcs?channel=0&stream=0&token=…
[Reolink Doorbell] Sending play command with path: channel0_sub.bcs?channel=0&stream=0&token=…
[Reolink Doorbell] Sending setBufferLength...
[Reolink Doorbell] Starting to yield video/audio packets...
[Reolink Doorbell] Socket received FIN          ← ~150 ms after getStreamLength
[Reolink Doorbell] RTMP sub.bcs prebuffer session ended
[Reolink Doorbell] Reolink Doorbell restarting prebuffer session in 5 seconds
```

Streaming the same URL from the same host with `ffmpeg`/`ffprobe`/`rtmpdump` works indefinitely.

## Root cause

Scrypted's custom RTMP client sends `getStreamLength` between `createStream` and `play`. Reolink's RTMP server appears to either treat this as an invalid command for a live stream or simply not implement it, and responds by tearing down the TCP connection. By the time the `play` command goes out (the very next message), the camera has already started sending FIN.

A wire-capture-confirmable hint is that the comment claims to match ffmpeg, but `libavformat`'s RTMP client (`libavformat/rtmpproto.c`, function `gen_get_stream_length`) only sends `getStreamLength` when the server has advertised a seekable/VOD stream — never as a default before live playback. The current Scrypted client sends it unconditionally.

## Fix

Remove the unconditional `getStreamLength` send (keep URL parsing, `sendPlay`, and `setBufferLength`):

```diff
-    // Send getStreamLength then play (matching ffmpeg's order)
+    // Compute playPath and start playback
     const parsedUrl = new URL(this.url);
     // Extract stream name (after /app/)
     const parts = parsedUrl.pathname.split('/');
     const streamName = parts.length > 2 ? parts.slice(2).join('/') : '';
     const playPath = streamName + parsedUrl.search;

-    this.console?.log('Sending getStreamLength with path:', playPath);
-    const getStreamLengthData = this.encodeAMF0Command('getStreamLength', this.transactionId++, null, playPath);
-    this.sendMessage(5, 0, RtmpMessageType.COMMAND_AMF0, 0, getStreamLengthData);
-
     this.console?.log('Sending play command with path:', playPath);
     this.sendPlay(this.streamId, playPath);
```

## Verification

Applied this exact patch to the deployed bundle on my Scrypted instance. Before the patch: ~11 prebuffer failures per minute (a FIN within ~150 ms of every play). After the patch: **0 FINs** over a 10-minute observation window, a single long-lived prebuffer session per stream, and HomeKit/Frigate consumers receive video normally. Other camera vendors (Hikvision, Amcrest, ONVIF-via-RTMP) that I do not own should be regression-tested by the maintainer or other users since their servers may behave differently if they actually rely on `getStreamLength` being absent or present.

I'll open a PR with this diff.

thllxb:
Fixes #2055.

## Summary

`RtmpClient.setup()` in `plugins/prebuffer-mixin/src/rtmp-client.ts` currently sends an AMF `getStreamLength` command between `createStream` and `play`. Some live RTMP servers — confirmed on Reolink cameras (e.g. Reolink Video Doorbell WiFi, firmware `v3.0.0.4662_2508071282`) — respond by closing the TCP connection (FIN) ~100–150 ms later, before the `play` command can take effect. The rebroadcast prebuffer then enters a tight ~5-second reconnect loop and never produces a usable stream. See #2055 for full reproduction and log evidence.

The original code comment claims to be "matching ffmpeg's order", but ffmpeg's `libavformat` RTMP client (`gen_get_stream_length` in `rtmpproto.c`) only sends `getStreamLength` conditionally for seekable/VOD streams — never as an unconditional preamble to `play` on a live stream. `rtmpdump` behaves the same way. The unconditional send in Scrypted is therefore both inaccurate to the stated intent and actively harmful to at least one popular camera vendor.

## Change

Removes only the three lines that build and send `getStreamLength`:

```diff
-    // Send getStreamLength then play (matching ffmpeg's order)
+    // Compute playPath from the URL (the part after /app/).
+    // Note: do NOT send a getStreamLength command here. ffmpeg's RTMP client only
+    // sends getStreamLength for seekable/VOD streams (libavformat rtmpproto.c
+    // gen_get_stream_length is guarded on RTMP_PT_*), never for live playback.
+    // Some live RTMP servers (e.g. Reolink cameras) FIN the TCP connection on
+    // receiving an unsolicited getStreamLength, which breaks the rebroadcast
+    // prebuffer instantly.
     const parsedUrl = new URL(this.url);
-    // Extract stream name (after /app/)
     const parts = parsedUrl.pathname.split('/');
     const streamName = parts.length > 2 ? parts.slice(2).join('/') : '';
     const playPath = streamName + parsedUrl.search;
 
-    this.console?.log('Sending getStreamLength with path:', playPath);
-    const getStreamLengthData = this.encodeAMF0Command('getStreamLength', this.transactionId++, null, playPath);
-    this.sendMessage(5, 0, RtmpMessageType.COMMAND_AMF0, 0, getStreamLengthData);
-
     this.console?.log('Sending play command with path:', playPath);
     this.sendPlay(this.streamId, playPath);
```

URL parsing, `sendPlay`, and `setBufferLength` are unchanged. `encodeAMF0Command` is still used by `sendConnect`, `sendCreateStream`, and `sendPlay`.

## Verification

Tested on a Reolink Video Doorbell WiFi (`D340W`, firmware `v3.0.0.4662_2508071282`) running against Scrypted `latest` (Docker, with the bundled `@scrypted/prebuffer-mixin@0.10.66`).

Before the change:
- `~11` prebuffer failures per minute on `RTMP sub.bcs`.
- Every `Sending play command` log line is followed by `Socket received FIN` within ~150 ms.
- Rebroadcast never delivers a frame to HomeKit.

After the change (verified with the equivalent surgical patch applied to the deployed minified bundle, then restart of the container):
- `0` FIN events in a 10-minute observation window.
- A single long-lived prebuffer session.
- HomeKit cameras stream normally, and motion-detection events flow during the same session.

## Backwards compatibility

I do not own other RTMP camera brands. Since ffmpeg has always been able to play live RTMP without sending `getStreamLength`, no RTMP server should require it for `play`. If a particular server *does* require it (unlikely for the camera/IPC use-case this plugin targets), the previous behaviour can be restored without breaking the connection by sending `getStreamLength` *only after* receiving an `onStatus` reply to `play` rather than before.

### DIFF
--- a/plugins/prebuffer-mixin/src/rtmp-client.ts
+++ b/plugins/prebuffer-mixin/src/rtmp-client.ts
@@ -133,16 +133,17 @@ export class RtmpClient {
     }
     this.console?.log('Got createStream _result');
 
-    // Send getStreamLength then play (matching ffmpeg's order)
+    // Compute playPath from the URL (the part after /app/).
+    // Note: do NOT send a getStreamLength command here. ffmpeg's RTMP client only
+    // sends getStreamLength for seekable/VOD streams (libavformat rtmpproto.c
+    // gen_get_stream_length is guarded on RTMP_PT_*), never for live playback.
+    // Some live RTMP servers (e.g. Reolink cameras) FIN the TCP connection on
+    // receiving an unsolicited getStreamLength, which breaks the rebroadcast
+    // prebuffer instantly.
     const parsedUrl = new URL(this.url);
-    // Extract stream name (after /app/)
     const parts = parsedUrl.pathname.split('/');
     const streamName = parts.length > 2 ? parts.slice(2).join('/') : '';
     const playPath = streamName + parsedUrl.search;
-
-    this.console?.log('Sending getStreamLength with path:', playPath);
-    const getStreamLengthData = this.encodeAMF0Command('getStreamLength', this.transactionId++, null, playPath);
-    this.sendMessage(5, 0, RtmpMessageType.COMMAND_AMF0, 0, getStreamLengthData);
 
     this.console?.log('Sending play command with path:', playPath);
     this.sendPlay(this.streamId, playPath);


koush:
This rtmp client is only really used with reolink anyways, but the session commands that are being sent by this client are verbatim what ffmpeg sends. FFmpeg is sending that get stream length from what I saw when man in the middling to implement RTMP.

do you have other reolinks?

thllxb:
> do you have other reolinks?

Unfortunately I don't have any other Reolink device.

Issue #1617, opened 2024-10-20
Loud interferenc noise on audio of Ring Video Doorbeel Pro 2


dklinger:
**Describe the bug**
Video Doorbell 2 Pro with homekit produces very weird and loud "interference noise" on audio.

**To Reproduce**
1. Setup Video Doorbell 2 Pro with Scrypted. 
2. Enable Homekit Plugin
3. Set Homekit RTP sender to ffmpeg ("Scrypted" mode does send no audio at all)
4. Get very weird and loud "interference noise" on audio when viewing a video.

**Expected behavior**
No interference noise. I tried "Transcode Audio" but without luck. Same weird interference noise.

**Server (please complete the following information):**
 - Installation Method: Docker

**Client (please complete the following information, if applicable):**
 - Software: iOS Home App

Pinkytoys1:
Hello everyone, I have the same problem! 
Anyone else?

Pinkytoys1:
Anyone ?

pierrevanmever:
I have also the exact same issue. 
When viewing HKSV (recorded after motion) I can see video, but audio is mangled, you can't hear anything
Live view is working, when I press the talk button, audio is coming from the Ring doorbell, however no audio is going back to Homekit. 
When viewing the recorded clips in scripted, audio works.

So in short:
In live view, no audio is coming in, when clips are saved in HKSV, audio is mangled, when viewing clips directly in scripted, audio works.

I've ordered the new generation of the doorbell pro to test out, to see if this solves the issues, we got good return policies here in The Netherlands.

pierrevanmever:
Update: I tried the 'new' Doorbell Pro 2, 2024 edition, same issue.

alexjet:
I have the same issue - when viewing the recorded HKSV clips in Home app the sound kind of jumps from time to time, those distortions make hearing voice impossible.

bluefunoodle311:
I am encountering the same issue with two interior cameras. Audio in Scrypted is just fine, but when I view on Home app, it's popping in and out. I have two wired floodlight cameras and a wired doorbell that are working perfectly. Tried FFmpeg and Transcode Audio, but no change. Do I need to install/configure a separate FFmpeg plugin?

Restarted Ring Plugin. No change. Disabled and re-enabled the device in Homekit plugin and now I have no incoming audio at all. Pushing talk in Home app delivers voice perfectly over the camera. Video seems unaffected, with the exception of some random artifacts here and there on Home App.

[...]

PR #2063, opened 2026-06-10
feat(webrtc): add per-camera "Disable Opus Audio" option

laurent-builds:
## Problem
Some cameras — notably certain Ring Pro doorbells — deliver audio over their
direct WebRTC connection as 60ms SILK Opus that decodes to crackle on every
decoder tested (HomeKit/iOS, ffmpeg, browser), despite byte-perfect, in-order
RTP. Cameras that send 20ms CELT Opus are clean. The corruption is in the
camera's encoding on that path, upstream of Scrypted — which is why transcoding
downstream doesn't fix it.

## Fix
The audio is clean when the camera/media server transcodes it server-side
instead of forwarding raw Opus (Ring does this on a codec mismatch). This adds a
per-camera "Disable Opus Audio" option that offers only G.711 on the receiving
WebRTC connection, forcing that server-side transcode.

- Per-camera setting, default off — no change for existing cameras.
- When enabled: clean H.264 video + stable connection; audio drops to G.711.
- Fixes crackly audio in both live view and recordings.

## Testing
multiple Ring cameras: the affected doorbell goes crackly → clean with the toggle on;
other Ring cameras are unaffected and keep full 48kHz Opus with it off.

Related to #1617 — this matches the symptoms there (Ring doorbell audio crackle
over HomeKit). Would appreciate the reporter confirming whether enabling this
option resolves it on their device.

### DIFF
--- a/plugins/webrtc/src/main.ts
+++ b/plugins/webrtc/src/main.ts
@@ -37,7 +37,14 @@ function delayWorkerExit(f: ReturnType<WebRTCPlugin['createTrackedFork']>) {
 }
 
 class WebRTCMixin extends SettingsMixinDeviceBase<RTCSignalingClient & VideoCamera & RTCSignalingChannel & Intercom> implements RTCSignalingChannel, VideoCamera, Intercom {
-    storageSettings = new StorageSettings(this, {});
+    storageSettings = new StorageSettings(this, {
+        disableOpusAudio: {
+            title: 'Disable Opus Audio',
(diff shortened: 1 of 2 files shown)


koush:
This may just be a gap in the opus repacketizer https://github.com/koush/scrypted/blob/main/plugins/homekit/src/types/camera/opus-repacketizer.ts

koush:
https://github.com/koush/scrypted/blob/f7018ddd9f75fc5e6dca66b2613e3aa23290eca8/plugins/homekit/src/types/camera/opus-repacketizer.ts#L68

PR #2064, opened 2026-06-15
Add HomeKit diagnostics panel

nbeathoven:
Once the HomeKit plugin is running, it's basically a black box. When a camera stops recording to HomeKit Secure Video or a pairing won't show up in the Home app, there's nothing to look at  which mDNS advertiser it picked, what it bound to, whether the bridge published, which Apple device is connecting, or whether a recording session is even producing fragments. Figuring that out means SSHing in and reading log lines. This is mostly the case when you are on Docker with multi ethernet/Vlan networking.

This adds a read-only Diagnostics section to the plugin settings that puts it all in one place: bridge publish state, mDNS advertiser and bind addresses, connected HomeKit clients, active live streams and the last RTCP timeout, HKSV recording sessions with fragment counts, and how much disk the debug clips are using (they pile up fast when Save Recordings is left on). It's visibility only  nothing here changes pairing, streaming, or recording. The point is to be able to see what's happening before touching the parts that affect pairing.

### DIFF
--- a/plugins/homekit/README.md
+++ b/plugins/homekit/README.md
@@ -10,7 +10,7 @@ You can use the admin page provided by your camera manufacturer to configure cod
 
 The HomeKit plugin settings include a Diagnostics section with a read-only snapshot of the current bridge, network, live streaming, and HomeKit Secure Video state.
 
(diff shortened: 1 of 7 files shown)


koush:
this is too specific of a use case that really only serves a few users. and I don't want to adopt maintainer burden for this. the stuff you mentioned is all shown in the logs too.

the docs walk through HomeKit trouble shooting too.

PR #2069, opened 2026-06-21
reolink(nvr): add HTTPS option for the NVR/Hub api.cgi connection

### LINKED ISSUE #2071: Reolink Home Hub / Home Hub Pro can't be added — api.cgi is HTTPS-only
## Problem

The Reolink **Home Hub** and **Home Hub Pro** serve their `api.cgi` control endpoint over **HTTPS only** (typically port 443). The Reolink NVR/Hub client hardcodes `http://` for every `api.cgi` request, so one of these hubs can't be added or managed — the connection fails (e.g. `EPROTO` / "wrong version number" when an HTTP request reaches the HTTPS-only port).

## Environment

- Scrypted Reolink plugin, NVR/Hub mode
- Reolink Home Hub Pro (current firmware)

## Expected

Be able to add and manage a Home Hub (Pro) using its HTTPS `api.cgi` endpoint.

## Actual

Requests are sent as `http://` to the HTTPS-only endpoint, so the hub can't be added.

---

Addressed by #2069.

SnoElement:
Fixes #2071.

## Problem

The Reolink Home Hub and Home Hub Pro serve their `api.cgi` control endpoint over **HTTPS only** (typically port 443). The Reolink NVR client currently hardcodes `http://` for every `api.cgi` call, so these hubs can't be added or managed.

## Change

Adds an opt-in HTTPS control-path for the NVR/Hub `api.cgi` connection, preserving HTTP defaults for existing devices:

- A per-device **"Use HTTPS"** advanced setting (default **off**), plus the same toggle in the add-device flow.
- The scheme is threaded through `ReolinkNvrClient` as a proper `ReolinkProtocol = 'http' | 'https'` union and used by **every** `api.cgi` request — including the login/token bootstrap in `probe.ts` (`getLoginParameters` / `getDeviceInfo`), which previously always used `http://`.
- Scheme-aware port default: `httpPort || (https ? 443 : 80)`. The `httpPort` field no longer hard-defaults to 80; its `onPut` clears a stale other-scheme default (80↔443) so toggling HTTPS doesn't get stuck on `:80`, while a genuinely custom port is preserved.
- The HTTPS setting description is shared between the creator form and the device settings so the two can't drift.

## Validation

We made these changes to get the plugin working on our own setup — a Reolink Home Hub Pro — and have confirmed success in day-to-day use:

- Verified on a real **Home Hub Pro**: hub-info fetch and login succeed over HTTPS (0 `EPROTO`).
- Backward compatibility verified end to end: the default stays HTTP, and the standalone camera path is untouched (HTTP-only).

## Note

The HTTPS option applies to the NVR/Hub flow; standalone cameras remain HTTP.

### DIFF
--- a/plugins/reolink/src/main.ts
+++ b/plugins/reolink/src/main.ts
@@ -9,7 +9,7 @@ import { OnvifIntercom } from './onvif-intercom';
 import { DevInfo } from './probe';
 import { AIState, Enc, isDeviceHomeHub, isDeviceNvr, ReolinkCameraClient } from './reolink-api';
 import { ReolinkNvrDevice } from './nvr/nvr';
-import { ReolinkNvrClient } from './nvr/api';
+import { HTTPS_API_SETTING_DESCRIPTION, ReolinkNvrClient } from './nvr/api';
 
 class ReolinkCameraSiren extends ScryptedDeviceBase implements OnOff {
     sirenTimeout: NodeJS.Timeout;
@@ -1280,6 +1280,13 @@ class ReolinkProvider extends RtspProvider {
             },
             {
                 subgroup: 'Advanced',
+                key: 'https',
+                title: 'Use HTTPS',
+                description: `Only applies when "Is NVR" is enabled. ${HTTPS_API_SETTING_DESCRIPTION}`,
+                type: 'boolean',
+            },
+            {
+                subgroup: 'Advanced',
                 key: 'skipValidate',
                 title: 'Skip Validation',
                 description: 'Add the device without verifying the credentials and network settings.',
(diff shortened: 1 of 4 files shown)


koush:
@apocaliss92

apocaliss92:
I don't think it's true that the hubs work only on https, it's just a setting. I would remove the comment stating this and also adding an additional setting for the https port, since both will still be available

koush:
> I don't think it's true that the hubs work only on https, it's just a setting. I would remove the comment stating this and also adding an additional setting for the https port, since both will still be available

yeah thats what I think too. can you confirm @SnoElement

PR #2078, opened 2026-06-25
twelvelabs: add Pegasus video understanding plugin

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

This adds an opt-in Twelve Labs Pegasus plugin (`@scrypted/twelvelabs`) that exposes a `ChatCompletion` device for video understanding.

## What it adds
- A new plugin under `plugins/twelvelabs` implementing `Settings` + `ChatCompletion`.
- Given a camera event clip (an `http(s)` video URL or a `data:video/...;base64,...` URL) plus a prompt, it calls the Twelve Labs Analyze (Pegasus) API and returns a natural-language description of what happened in the clip.
- A small, dependency-free REST client for the Analyze endpoint (`src/pegasus.ts`) and a pure request parser (`src/extract.ts`) that pulls the video + prompt out of an OpenAI-shaped chat completion body.

## Why it helps Scrypted
Any flow that can call a chat completion (NVR event pipelines, notifier automations, scripts) can now turn a motion/event clip into a description like "a delivery person left a package on the porch" instead of a generic "motion detected". It's the first `ChatCompletion` provider that is video-native rather than image/frame based.

## Opt-in / non-breaking
Fully opt-in: nothing runs until a user enters an API key in the plugin settings, and it changes no Scrypted defaults or existing behavior. No new dependencies are added to the build (the client uses the built-in `fetch`).

## How it was tested
- `npm run build` (scrypted-webpack) compiles the plugin cleanly.
- `npm test` (`node --test`) runs 9 no-network unit tests for the body builder, the request parser, and the client's response/error mapping, plus 1 live smoke test gated on `TWELVELABS_API_KEY` (skipped when unset).
- I ran the live test against the real Analyze endpoint with a key: auth and the request schema are accepted (verified end to end).

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### DIFF
--- a/plugins/twelvelabs/.gitignore
+++ b/plugins/twelvelabs/.gitignore
@@ -1,0 +1,5 @@
+.DS_Store
+out/
+node_modules/
+dist/
+.venv
--- a/plugins/twelvelabs/.npmignore
+++ b/plugins/twelvelabs/.npmignore
@@ -1,0 +1,9 @@
+.DS_Store
+out/
+node_modules/
+*.map
+src
+test
+.vscode
+dist/*.js
+.venv
--- a/plugins/twelvelabs/README.md
+++ b/plugins/twelvelabs/README.md
@@ -1,0 +1,39 @@
+# Twelve Labs Pegasus for Scrypted
+
+This plugin adds a [Twelve Labs](https://twelvelabs.io) Pegasus video-understanding
+device to Scrypted. Pegasus is a video-native model: given a video clip and a natural
+language prompt, it returns a text description of what happens in the clip.
+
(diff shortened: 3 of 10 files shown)


koush:
Cool, but no need for this to be in the monorepo, you are welcome to keep the plugin in your own org and maintain it there as well as publish it yourself.

mohit-twelvelabs:
Makes total sense, @koush — thanks for the quick look and the pointer. A standalone published plugin in our own org is a clean fit for Scrypted's architecture; we'll go that route. Appreciate it!

PR #2083, opened 2026-06-30
fix(alexa): v3 API compliance, fan speed, security panel, and camera streaming improvements

nberardi:
## Summary

Brings the Alexa Smart Home integration in line with the current v3 Smart Home API and improves camera streaming.

### API compliance, fan speed, security panel
- **BrightnessController**: report `brightness` under `Alexa.BrightnessController` instead of `Alexa.PowerController` so Set/AdjustBrightness responses match the advertised capability.
- **SmartVision ObjectDetection**: flatten the doubly-nested `payload.events` array.
- **SecurityPanelController**: add Arm/Disarm directive handlers (previously advertised but unhandled), wired to `armSecuritySystem`/`disarmSecuritySystem`. Enforce the spec rule rejecting a direct `ARMED_AWAY` → other-armed-state transition with `AUTHORIZATION_REQUIRED`. Drop the advertised `FOUR_DIGIT_PIN` auth that Scrypted can't honor.
- **Auth failures** now return an Alexa `INVALID_AUTHORIZATION_CREDENTIAL` ErrorResponse instead of a bare HTTP 500 that Alexa retries and surfaces as a generic skill error.
- **EndpointHealth**: downgrade to the documented v3.1 / connectivity-only shape; stop emitting the undocumented `battery` property (and `3.2` version) that risks the whole capability being rejected.
- **Fan**: add `Alexa.RangeController` (`Fan.Speed`, 0–100%) for devices implementing the Fan interface, alongside on/off.
- Cleanups: `addAccessToken` initializes `event.endpoint` as an object not an array; hoist `getArmState` to a shared export.

### Camera streaming
- **Resolution**: raise the proxied stream cap from 720p to 1080p. The old 1280×720 `screen` hint forced medium-resolution mode for every endpoint; capable displays (Echo Show 15, Fire TV) now render at 1080p while the transcoder still clamps width and auto-falls back to 720p when H.264 High isn't negotiated.
- **Two-way audio**: only advertise `isFullDuplexAudioSupported` when the camera implements the `Intercom` interface (webrtc emits `a=sendonly` for one-way cameras downstream).
- **TURN**: stop unconditionally disabling TURN and expose a "Use TURN Servers" plugin setting (on by default). Alexa sessions are proxied and often cross NATs where a TURN relay is the only path; with trickle ICE disabled (per Alexa's spec, all candidates must be in the initial SDP), omitting the relay candidate left NAT-blocked sessions with no fallback. Defaults defer to the WebRTC plugin's setting, matching the Google Home integration.

## Testing
Deployed to a local Scrypted server and exercised: light brightness, fan speed, security panel arm/disarm, camera streaming on one-way and two-way cameras.

## Notes
An async (event-gateway) answer for `InitiateSessionWithOffer` to avoid the 6s synchronous-response deadline on cold cameras is a known follow-up, not included here.

### DIFF
--- a/plugins/alexa/package.json
+++ b/plugins/alexa/package.json
@@ -1,6 +1,6 @@
 {
    "name": "@scrypted/alexa",
-   "version": "0.4.0",
+   "version": "0.5.0",
    "scripts": {
       "scrypted-setup-project": "scrypted-setup-project",
       "prescrypted-setup-project": "scrypted-package-json",
--- a/plugins/alexa/package.json
+++ b/plugins/alexa/package.json
@@ -1,6 +1,6 @@
 {
    "name": "@scrypted/alexa",
-   "version": "0.3.7",
+   "version": "0.4.0",
    "scripts": {
       "scrypted-setup-project": "scrypted-setup-project",
       "prescrypted-setup-project": "scrypted-package-json",
--- a/plugins/alexa/src/common.ts
+++ b/plugins/alexa/src/common.ts
@@ -1,4 +1,4 @@
-import { Battery, Online, PowerSensor, ScryptedDevice, ScryptedInterface, HttpResponse } from "@scrypted/sdk";
+import { Online, ScryptedDevice, ScryptedInterface, HttpResponse } from "@scrypted/sdk";
 import { v4 as createMessageId } from 'uuid';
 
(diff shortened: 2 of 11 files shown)


koush:
great, you have publish access on npm?

nberardi:
published

PR #2098, opened 2026-07-18
fix(homekit): crackling and muted Opus audio from Ring doorbells (#1617)

laurent-builds:
Ring doorbells send Opus audio in SILK mode with 60ms frames, encoded above 0 dBFS. This causes two bugs:

- **Crackling**: ffmpeg's default Opus decoder doesn't apply the soft-clip the Opus spec expects for content above 0 dBFS, so it hard-clips instead — audible as crackle. This affected recordings and any transcoded live stream.
- **Muted audio**: 60ms frames can't be repacketized into HomeKit's negotiated 20ms packet time, so codec-copied live audio gets muted.

**Fix:**
- Recordings and audio transcodes now decode Opus with `libopus`, which soft-clips correctly.
- Live streaming checks each packet's Opus mode: CELT streams (the common case) are copied through unchanged. SILK or long-frame streams are transcoded through libopus with a small gain margin and re-packetized to the correct duration, falling back to copy if the transcoder fails to start.
- Fixed the opus repacketizer to use each packet's actual frame duration instead of assuming 20ms, plus two related RFC 6716 length-encoding bugs.

Tested on a Ring Video Doorbell Pro (HomeKit + HKSV) — live view and recordings are now clean. Verified CELT cameras are unaffected (no transcoding, unchanged behavior).

Supersedes #2063, which worked around this by disabling Opus entirely (forcing G.711). This fixes the root cause with no toggle and keeps full Opus audio quality.

### DIFF
--- a/plugins/homekit/src/types/camera/camera-streaming-ffmpeg.ts
+++ b/plugins/homekit/src/types/camera/camera-streaming-ffmpeg.ts
@@ -1,14 +1,20 @@
 import { RtpPacket } from '@koush/werift-src/packages/rtp/src/rtp/rtp';
 import { getDebugModeH264EncoderArgs } from '@scrypted/common/src/ffmpeg-hardware-acceleration';
 import { addVideoFilterArguments } from '@scrypted/common/src/ffmpeg-helpers';
-import { createBindZero } from '@scrypted/common/src/listen-cluster';
+import { closeQuiet, createBindZero, reserveUdpPort } from '@scrypted/common/src/listen-cluster';
+import { ffmpegLogInitialOutput, safeKillFFmpeg } from '@scrypted/common/src/media-helpers';
 import { getSpsPps } from '@scrypted/common/src/sdp-utils';
-import { FFmpegInput, MediaStreamDestination, ScryptedDevice, VideoCamera } from '@scrypted/sdk';
+import sdk, { FFmpegInput, MediaStreamDestination, ScryptedDevice, VideoCamera } from '@scrypted/sdk';
+import child_process from 'child_process';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
 import { RtpTrack, RtpTracks, startRtpForwarderProcess } from '../../../../webrtc/src/rtp-forwarders';
 import { AudioStreamingCodecType, SRTPCryptoSuites } from '../../hap';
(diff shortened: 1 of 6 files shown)


plugins/homekit/src/types/camera/camera-streaming-ffmpeg.ts
```diff
@@ -181,16 +192,147 @@ export async function startCameraStreamFfmpeg(device: ScryptedDevice & VideoCame
...
                     audioSampleRate: session.startRequest.audio.sample_rate,
-                    framesPerPacket: opusFramesPerPacket,
                 }
             );
 
             firstPacket = function () {
                 audioSender.sendRtcp();
             };
 
+            // SILK and Hybrid mode Opus (TOC config < 16, e.g. from Ring doorbells) routinely
```
koush, line 202: transcoding concerns shouldn't be in the HomeKit plugin. secondly, this seems like a lot of steps with what could be achieved by short circuiting the opus copy and forcing the transcode and letting the already existing audio transcode process handle it?

koush:
I'm also confused why this is attributed to the opus ffmpeg decoder? opus to opus does not utilize ffmpeg at all if I'm remembering correctly, its a straight stream copy/repacketization, and isn't touched by ffmpeg at all. I haven't seen any logs for this issue. a log of the issue of the without any patches would clear up these questions.

was this an ai generated submission?

plugins/homekit/src/types/camera/camera-streaming-ffmpeg.ts
laurent-builds: Agreed, it shouldn't live here. I didn't force the existing transcode path because codecCopy is decided before any packets arrive, and the frame duration isn't in mediaStreamOptions, so you can't tell 60ms from 20ms until the first packet's TOC. Options: always transcode opus, or expose frame duration in the stream metadata so the decision happens up front. Can rework whichever you prefer.

koush:
I think two separate pull requests is warranted here. the recording fix doesn't look like it can be merged as is, but if the opus to opus is just a matter of fixing the repacketizer, I can merge that.

koush:
need log

koush:
no this is not sufficient, see the bug report instructions

PR #2113, opened 2026-08-05
reolink: add per-camera Motion Trigger setting (Motion + AI / AI Only)

marl1w:
The motion sensor was set by two independent 1 Hz pollers with equal weight: GetMdState and the AI object detection state. HomeKit Secure Video has only a MOTION recording trigger, so any raw motion detection became an HKSV clip and a notification.

On an NVR-backed install that makes the camera notify on things that are not subjects at all. A day/night IR filter switch changes the whole frame and trips motion detection even at the lowest sensitivity, as does a floodlight turning on, and both recur every single day at dusk and dawn.

Motion Trigger gates the raw motion detector without touching the AI path, so person, vehicle and animal detections still drive HomeKit. It defaults to Motion + AI, preserving existing behaviour.

The PIR branch is deliberately exempt: a battery camera has no AI state, so PIR is its only trigger and gating it would leave the camera with none.

Also make the AI poll survive a bad response. It previously returned out of the loop for good when a poll reported no supported classes, and for non-NVR devices on the first exception, so a camera that was briefly unreachable never detected objects again. That was masked by raw motion keeping the sensor alive; with AI Only there is no such fallback, and a camera would have gone permanently silent with nothing indicating why. Failures now back off exponentially to 60s instead of exiting.

### DIFF
--- a/.github/workflows/upstream-sync.yaml
+++ b/.github/workflows/upstream-sync.yaml
@@ -6,12 +6,18 @@ name: Upstream sync
 # Notes:
 #   * Scheduled workflows are disabled by default in forks - enable Actions on
 #     the "Actions" tab once, otherwise the cron never fires.
-#   * With the default GITHUB_TOKEN, pushes made here do NOT trigger other
-#     workflows, so `test.yml` will not run on the sync PR. To get CI on it,
-#     create a PAT with `repo` scope, store it as the `SYNC_TOKEN` secret, and
-#     this workflow will use it automatically.
+#   * A `SYNC_TOKEN` secret is effectively required. The default GITHUB_TOKEN
+#     may never push a commit that touches `.github/workflows/**` (no
+#     `permissions:` key can grant this), and upstream regularly changes its
+#     own workflows, so those syncs are rejected outright. It also cannot
+#     trigger other workflows, so `test.yml` would not run on the sync PR.
+#     Create a PAT - classic with `repo` + `workflow` scopes, or fine-grained
+#     with Contents and Workflows read/write - and store it as `SYNC_TOKEN`.
 #   * Submodules are deliberately not checked out: the merge only touches the
 #     superproject, and cloning them would make every run much slower.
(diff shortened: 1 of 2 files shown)


koush:
can merge the poll recovery, but the change to only trigger motion on upstream AI events is incorrect and won't be merged.

marl1w:
> can merge the poll recovery, but the change to only trigger motion on upstream AI events is incorrect and won't be merged.

Thanks for the feedback, @koush.

Could you clarify what is incorrect about allowing upstream AI events to be used as the motion trigger?

In my testing, Reolink’s raw motion detection produces frequent (20/30+ per day even after a fine tuning of the motion detection setting) false positives from conditions such as fog, rain, moving shadows, IR mode changes, and lighting changes. The goal of the optional “AI Only” setting is to prevent those raw events from triggering downstream systems such as HKSV, while still allowing person, vehicle, and animal detections reported by the camera to trigger them.

The default remains “Motion + AI,” so the existing behavior is preserved unless a user explicitly changes the setting.

I understand that this may not fit Scrypted’s intended motion-event model or architecture. Could you suggest the preferred way to handle cameras that continuously generate unreliable raw motion events? I would be happy to revise the implementation accordingly.

koush:
homekit will handle the motion correctly downstream. gating it on the native camera detection means HomeKit will have more false negatives.

marl1w:
> homekit will handle the motion correctly downstream. gating it on the native camera detection means HomeKit will have more false negatives.

Gotcha, but the use case is when HKSV recording is disabled. In that configuration, HomeKit does not analyze video downstream; it simply trusts the exposed motion sensor and sends a “motion detected” notification. Live testing shows Reolink’s raw motion state can produce repeated notifications when no visible motion occurred.

The optional AI-only mode addresses those false positives for users who prefer notification accuracy and knowingly accept the risk of additional false negatives. This also cause the existing raw-motion behavior would remain the default.

Issue #2122, opened 2026-08-15
Amcrest Digest authentication fails because qop and nc are incorrectly quoted.  Suggested Fix Provided.


jparchem:
** Before You Submit**

- [X] I checked that my issue isn't already filed: [Search open issues](https://github.com/koush/scrypted/issues).   
(Note:  The issue was first reported in issue [#1420](https://github.com/koush/scrypted/issues/1420) but that issue was closed without resolution)
- [X] I checked the relevant camera/device and/or plugin `Log` in the `Management Console` for errors or warnings that may help identify and resolve the issue myself.

## Describe the bug

On specific Amcrest camera models, the Amcrest plugin fails to authenticate because Scrypted generates a non-compliant HTTP Digest Authorization header.  This bug was first reported in issue #1420, but no resolution was found and that issue marked closed.  Here I have identified the root cause, and provided a suggested path to remediation.

I have submitted a correspond PR #2123 for your review

### Details

The camera’s initial Digest challenge is handled correctly, and Scrypted does retry the request. However, the authenticated retry serializes these fields as quoted strings:

```http
qop="auth"
nc="00000001"
```

[RFC 7616  Section 3.4](https://httpwg.org/specs/rfc7616.html#rfc.section.3.4 ) stipulates that both values are to be **unquoted** tokens:

> “For historical reasons, a sender MUST NOT generate the quoted string syntax for the following parameters: algorithm, qop, and nc.”

```http
qop=auth
nc=00000001
```

Most of my Amcrest models appear to tolerate the quoted form (e.g. `IP8M-2496E-V2`, `IP5M-B1276EW-AI`, `IP4M-1041W`, etc.) which may explain why this issue is model-dependent. Other models like my `IP5M-1190EW` (and other cameras models such as those mentioned in Issue 1420) use firmware that strictly enforce it and reject the quoted form with `401 Unauthorized`.

## Environment

- Scrypted: `0.141.0`
- Amcrest plugin: `0.0.168`
- Affected camera: Amcrest `IP5M-1190EW`
- Authentication: HTTP Digest with `qop="auth"` and implicit MD5

## Symptoms

The affected device repeatedly logs:

```text
error retrieving stream configurations Error: http response statusCode 401
listen loop connection failed, restarting listener. http response statusCode 401
```

## Relevant stack frames:

```text
@scrypted/server/src/fetch/index.ts:45
@scrypted/server/src/fetch/http-fetch.ts:129
[...]

Issue #1420, opened 2024-04-05
Amcrest authentication issues


mrala:
**Describe the bug**
I'm receiving 401 errors using the Amcrest plugin on a `IPC-T2431T-AS S2` camera (no problems on the older non-S2 revision `IPC-T2431T-AS`). It's attempting to use digest auth but getting 401 responses. However, I'm able to successfully curl the camera endpoints with digest auth. The RTSP video feed is working correctly. I've configured security settings on the `S2` variant identically  as an older camera. This is breaking event/motion detection.

**To Reproduce**
1. Add a `IPC-T2431T-AS S2` camera using the Amcrest plugin
2. View the console for the Amcrest plugin
3. See example errors:
    ```
    error retrieving stream configurations Error: http response statusCode 401
        at checkStatus (C:\server\src\fetch\index.ts:43:15)
        at httpFetch (C:\server\src\fetch\http-fetch.ts:124:28)
        at processTicksAndRejections (/<snip>/plugins/amcrest/out/lib/internal/process/task_queues.js:95:5)
        at AmcrestCameraClient.request (C:\src\amcrest-api.ts:126:26)
        at C:\src\main.ts:422:41 {stack: 'Error: http response statusCode 401
        at ch…t-api.ts:126:26)
        at C:\src\main.ts:422:41', message: 'http response statusCode 401'}
    preparing event listener http://192.168.0.5:80/cgi-bin/eventManager.cgi?action=attach&codes=[All]
    listen loop connection failed, restarting listener. http response statusCode 401
    ```
    
    I'm able to use digest auth in a curl request. The camera reports a login failure.
    ```
    $ curl --user "<snip>:<snip>" "http://<snip>:80/cgi-bin/eventManager.cgi?action=attach&codes=\[All\]" --digest
    --myboundary
    Content-Type: text/plain
    Content-Length: 183
    
    Code=LoginFailure;action=Pulse;index=0;data={
            "Name": "admin",
            "Type": "Web3.0",
            "Address":      "<scrypted-ip-address>",
            "Error":        4,
            "LocaleTime":   "2024-04-04 22:41:04",
            "UTC":  1712284864
    }
    --myboundary
    Content-Type: text/plain
    Content-Length: 119
    
    Code=VideoMotion;action=Start;index=0;data={
            "LocaleTime":   "2024-04-04 22:41:47",
            "UTC":  1712284907,
            "Name": "IPC"
    }
    
    --myboundary
    Content-Type: text/plain
[...]

PR #2123, opened 2026-08-16
auth-fetch: fix Digest authentication token quoting

jparchem:
## Summary

Fix HTTP Digest authentication headers so `algorithm`, `qop`, and `nc` use unquoted token syntax.  Reported in issue #2122 

## Root cause

`http-auth-utils` quotes every Digest parameter when building the `Authorization` header. RFC 7616 section 3.4 requires clients not to quote `algorithm`, `qop`, or `nc`.

Some Amcrest camera models tolerate the quoted form, but others reject the authenticated retry with another HTTP 401.

## Validation

- Captured the expected initial Digest challenge from an affected Amcrest camera.
- Reproduced `401 -> 401` with quoted `qop` and `nc`.
- Repeated the same request with identical Digest inputs and unquoted tokens, producing `401 -> 200`.
- Verified `curl --digest` succeeds against the same endpoint.
- Applied the equivalent change as a temporary bundle canary in Scrypted; codec discovery, event listening, snapshots, and RTSP setup recovered without further authentication failures.

## Testing

The `@scrypted/auth-fetch` package does not currently have an automated test command (its test script exits with "Error: no test specified").

### DIFF
--- a/packages/auth-fetch/src/auth-fetch.ts
+++ b/packages/auth-fetch/src/auth-fetch.ts
@@ -55,7 +55,11 @@ async function getAuth(options: AuthFetchOptions, url: string | URL, method: str
             ...digest.data,
         });
 
-        return header;
+        // RFC 7616 requires algorithm, qop, and nc to be unquoted tokens.
+        return header
+            .replace(`algorithm="${digest.data.algorithm}"`, `algorithm=${digest.data.algorithm}`)
+            .replace(`qop="${digest.data.qop}"`, `qop=${digest.data.qop}`)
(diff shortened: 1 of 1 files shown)


keith721:
Subscribed.  One critical Dahua camera (EmpireTech IPC-T221R-ZAS-S3) cannot be added to Scrypted plug-in in Home Assistant because of this authentication issue.  Previously reported on https://github.com/koush/scrypted/issues/1420#issuecomment-5023660380

koush:
https://github.com/koush/scrypted/commit/bd451e71ecec5eb0ccb10b9114fd0cdaf3a0ac63

Issue #1793, opened 2025-04-21
HomeKit - Smart detections only


herbi3:
G'day! Any chance to have Scrypted to only send smartdetections to HomeKit?

Idea is, HomeKit takes too long to process smart detections, quite often long after the event.

However, if general motion events were able to be disabled via Scrypted and only smartdetections from UniFi were sent to HomeKit, then we can set HomeKit notifications to "any motion"

I have tried the smart detection sensor with Scrypted, while that does create a seperate sensor, it is not linked to the camera, thus no notifications via HomeKit.

herbi3:
Any feedback for this one?

PR #2126, opened 2026-08-18
unifi-protect: optionally trigger motion from smart detections

momouiui:
## Problem

Some UniFi Protect cameras can report `smartDetectZone`/`smartDetectLine` events without reporting a regular `motion` event. The plugin currently publishes those events through `ObjectDetector`, but the camera's `MotionSensor` remains inactive. Motion-based consumers such as HomeKit Secure Video therefore do not start a recording for those detections.

This was observed with a G5 PTZ, but the behavior should not be keyed to a camera name or model.

## Solution

- Add an opt-in per-camera advanced setting, **Smart Detections Trigger Motion**.
- When enabled, route `smartDetectZone` and `smartDetectLine` through the existing `debounceMotionDetected` path.
- Keep the setting disabled by default, so existing behavior is unchanged for every camera.
- Keep regular `motion` events working exactly as before and ignore unrelated event types.

This avoids model/name heuristics and lets users enable the fallback only on cameras that need it. Cameras that already report regular motion can leave the setting disabled.

## Tests

- Added focused tests for regular motion, enabled/disabled smart detection fallback, unrelated events, and the actual event-to-debounce routing.
- Confirmed the smart-detection test fails when the fallback implementation is temporarily reverted to the previous motion-only behavior.
- `npm test` — 4 tests passed on Node 18 and Node 22.
- `npm run build` — passed for `plugins/unifi-protect` after the repository's standard `npm-install.sh` setup. The build reports the existing `server/src/rpc.ts` TS4113 warning but completes successfully.

Related context: #1793

### DIFF
--- a/plugins/unifi-protect/package-lock.json
+++ b/plugins/unifi-protect/package-lock.json
@@ -17,7 +17,9 @@
          },
          "devDependencies": {
             "@types/node": "^22.15.29",
-            "@types/ws": "^8.18.1"
+            "@types/ws": "^8.18.1",
+            "ts-node": "^10.9.2",
+            "typescript": "^5.9.3"
          }
       },
       "../../common": {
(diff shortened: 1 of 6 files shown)


koush:
this comes up often, and no. that gates apple's detections behind unifi's less accurate ones.

PR #2127, opened 2026-08-20
reolink: add opt-in Baichuan protocol for lower-latency two-way audio

Alberto115x:
## Summary

Adds an opt-in, lower-latency alternative to the ONVIF intercom backchannel for Reolink devices, using Reolink's proprietary "Baichuan" binary protocol (TCP port 9000) instead of ONVIF/RTP for two-way audio.

**Motivation**: on a Reolink Video Doorbell PoE, the existing ONVIF backchannel had noticeably high two-way-audio latency even on pure LAN. Baichuan sends audio directly with no RTP/jitter buffering, which measured significantly lower latency on the real device.

**Scope**: audio only. This is purely additive — ONVIF remains the default and is completely untouched, including `onvif-events.ts` object/motion detection, which does not use the intercom path at all and is unaffected. Video over Baichuan was considered but deliberately left out of scope for this PR.

## What's new

- `plugins/reolink/src/baichuan/{crypto,protocol,adpcm,xml,client}.ts` — Baichuan protocol implementation: AES/MD5-based login and message encryption, binary message framing, ADPCM audio encode/decode, minimal XML command builder, and a client handling login + talk-session lifecycle.
- `plugins/reolink/src/baichuan-intercom.ts` — intercom implementation wired to the existing `Intercom` interface, mirroring the shape of `onvif-intercom.ts`.
- `plugins/reolink/src/main.ts` — minimal additive wiring: a new `useBaichuanTwoWayAudio` StorageSettings toggle (**default off**) that, when enabled, routes `startIntercom`/`stopIntercom` through `BaichuanIntercom` instead of `OnvifIntercom`; a `baichuanPort` advanced override (defaults to 9000). No existing ONVIF code paths are modified.

## Validation

- Live-tested end-to-end against a real Reolink Video Doorbell PoE (login, talk-session setup, real-time ADPCM audio streaming).
- Protocol details (message IDs, encryption, framing) cross-checked against AlexxIT/go2rtc PR #2263 for consistency with another independent implementation.
- Iterated through a few robustness fixes found only under live production use: correct Baichuan message IDs (`TalkConfig=201`, not `11`), a time-based (8s) reply-drain instead of a fixed message-count skip, persistent socket error/timeout handlers (a missing handler could previously crash the whole Scrypted process), real-time ADPCM pacing, and not logging the expected caller-hangup race as an error.

## Out of scope / follow-ups

- Video over Baichuan (Phase 2) — not started, left for a future PR if there's interest.
- This is opt-in and defaults off specifically so it has zero behavioral impact on existing ONVIF users unless they explicitly enable it.

### DIFF
--- a/plugins/reolink/package-lock.json
+++ b/plugins/reolink/package-lock.json
@@ -11,7 +11,9 @@
          "dependencies": {
             "@scrypted/common": "file:../../common",
             "@scrypted/sdk": "file:../../sdk",
-            "onvif": "file:../onvif/onvif"
+            "@types/xml2js": "^0.4.14",
+            "onvif": "file:../onvif/onvif",
+            "xml2js": "^0.6.2"
          },
          "devDependencies": {
             "@scrypted/sdk": "file:../../sdk",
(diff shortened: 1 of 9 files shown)


koush:
@apocaliss92 already did this in the reolink native plugin.

PR #2132, opened 2026-09-03
Add UniFi Protect API Key Only connection mode

nickberardi:
## Summary

Adds an **API Key Only** connection mode to the UniFi Protect plugin, alongside the existing local user account flow — modeled after Home Assistant’s approach in [home-assistant/core#176410](https://github.com/home-assistant/core/pull/176410).

### Connection modes

| Mode | Auth | Scope |
|------|------|--------|
| **Local User** (default) | Username / password against the private Protect API | Full feature set (unchanged) |
| **API Key Only** | UniFi OS Integration API key (`X-API-KEY`) against the public Integration API | Cameras + lights |

API keys are created in the UniFi OS local portal under **Settings → Control Plane → Integrations**.

### API Key Only capabilities

- Cameras: RTSPS streams, snapshots, motion/smart detections, doorbell ring, status LED
- Lights: on/off, brightness, PIR motion
- Public events/devices websockets for realtime updates

### Not available in API Key Only (use Local User)

- Sensors and locks
- Two-way audio / intercom
- Privacy masks, dynamic bitrate, optical zoom
- Fingerprint sensors and package-camera snapshots

### Implementation notes

- New `ProtectPublicApi` client (`plugins/unifi-protect/src/public-api.ts`) talks to `/proxy/protect/integration/v1/*`
- Public camera/light payloads are adapted into bootstrap-compatible shapes so discovery and device helpers can be shared
- Settings UI shows username/password or API key based on the selected mode
- Existing Local User installs are unaffected (default mode)
- Public API client rate-limits to UniFi’s 10 req/s window, retries 429s, and safely resets CONNECTING websockets
- Version bumped to `0.2.0`

## Test plan

- [x] API Key Only connects with IP + API key against Protect 7.2.x
- [x] Fixed 429 / websocket uncaughtException loop on bootstrap (rate limit + safe WS reset)
- [x] Local User username/password path still connects and streams
- [x] Discover/adopt cameras; verify RTSPS playback and snapshots
- [x] Motion / smart detect / doorbell ring events update Scrypted sensors
- [ ] Lights on/off and brightness work
- [ ] Sensors/locks are not offered in API Key Only mode
- [ ] Switching back to Local User still works with stored credentials

### DIFF
--- a/plugins/unifi-protect/README.md
+++ b/plugins/unifi-protect/README.md
@@ -1,16 +1,41 @@
 # Unifi Protect Plugin for Scrypted
 
-The Unifi Protect Plugin connects your Unifi Cameras to Scrypted. 
+The Unifi Protect Plugin connects your Unifi Cameras to Scrypted.
 
-## Requirements
+## Connection modes
+
+The plugin supports two connection modes:
+
+### Local User (default)
+
+Uses a Protect local user account against the private Protect API. This provides the full feature set (cameras, lights, sensors, locks, two-way audio, privacy masks, dynamic bitrate, etc.).
+
+Requirements:
 * The Protect appliance such as a Cloud Key or Dream Machine
 * Protect user account with **Local Administrator** permissions. **NOTE**: This may be downgraded to Read Only under certain situations (see `Troubleshooting`)
    * Two Factor Authentication will not work.
    * A local account is recommended in case the Ubiquiti SSO service goes down.
 
+### API Key Only
+
+Uses a UniFi OS Integration API key with no local user account — the same approach Home Assistant added for UniFi Protect. Authentication is via the `X-API-KEY` header against Protect's public Integration API.
+
+Create a key in the UniFi OS local portal: **Settings → Control Plane → Integrations**.
+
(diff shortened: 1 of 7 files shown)


koush:
I think this is fine to have sitting around in case they ever break their undocumented API, but it's generally unnecessary imo. too much breakage. its more compelling if they actually offer everything their undocumented api supports that we need.

image

nberardi:
The reason I created this was primarily selfish:

1) The UniFi Fabric Identity SSO was making it increasingly complex to manage and create local only accounts. 

2) The private API doesn’t work with Fabric managed users. 

2) HA recently moved in a similar [dual support direction](https://www.home-assistant.io/integrations/unifiprotect/) in their 2026.9 release, and I wanted to stay pace with the API changes for Scrypted. 

In regards to your comments, yes the public Integration API still doesn’t cover everything Scrypted needs, primarily PTZ, so Local User remains the full-featured default in the configuration. 

This PR keeps that path unchanged and adds API Key Only as an optional mode for the subset the public API does support today (steams, snapshots, lights).

